PR #2, opened 2020-06-25
add html form to profile page

alikew24:
added an html form, a nav bar, and some designs to the profile page.


src/main/webapp/profile.html
```diff
@@ -3,11 +3,90 @@
...
+          <div>
+            <input type="file" accept="image/*" name="photo">
+          </div>
+          <div class="form-group">
+            <label for="name">Name:</label>
+            <input type="text" class="form-control" placeholder="Enter name" id="name" width="50">
+          </div>
+          <div class="form-group">
+            <label for="bio">Bio (max 300 characters):</label>
+            <textarea name="bio" class="form-control" placeholder="Enter a couple of words about yourself" id="bio" maxlength="300"></textarea>
```
jessdejong, line 50: nit: line wrap this line to 100-120 characters

src/main/webapp/style.css
```diff
@@ -0,0 +1,42 @@
...
+    background-color: #303c6c;
+    color: white;
+}
+.nav-text-style {
+    color: #303c6c;
+    font-size: 30px;
+}
+#name {
+    align-self: center;
+    align-content: center;
```
jessdejong, line 25: nit: alphabetize css properties

src/main/webapp/style.css
```diff
@@ -0,0 +1,42 @@
...
+    color: #303c6c;
+    text-align: center;
+}
+.button-background {
+    background-color: #303c6c;
+    color: white;
+}
+.nav-text-style {
+    color: #303c6c;
+    font-size: 30px;
```
jessdejong, line 21: Instead of using `px` for the `font-size`, a scaling unit like `em` would be better because it scales with the size of the screen.

alikew24:
thanks jess! added those edits

jessdejong:
@estep-google's turn, I think daniel hasn't accepted the collaborator request?

src/main/webapp/profile.html
```diff
@@ -3,11 +3,91 @@
...
+            <textarea name="bio" class="form-control" placeholder="Enter a couple of words about yourself" 
+            id="bio" maxlength="300"></textarea>
+          </div>
+          <div class="d-flex image-upload">
+            <div class="p-2">
+              <div>
+                <img src="images/placeholder.jpg" class="rounded-circle" alt="placeholder" width="200" height="200">
+              </div>
+              <div>
+                <input type="file" accept="image/*" name="photo">
```
estep-google, line 59: Mostly curiosity - why does each photo here have the same `name` field?

alikew24: Screenshot 2020-06-25 at 11 16 38 AM
this is what the page looks like! And each photo should not have the same name field i need to change that

src/main/webapp/profile.html
```diff
@@ -47,7 +47,8 @@ <h1>Profile</h1>
           </div>
           <div class="form-group">
             <label for="bio">Bio (max 300 characters):</label>
-            <textarea name="bio" class="form-control" placeholder="Enter a couple of words about yourself" id="bio" maxlength="300"></textarea>
+            <textarea name="bio" class="form-control" placeholder="Enter a couple of words about yourself" 
+            id="bio" maxlength="300"></textarea>
```
jessdejong, line 51: I think this line should be indented four spaces. https://google.github.io/styleguide/htmlcssguide.html#HTML_Line-Wrapping

PR #5, opened 2020-06-29
Add .class to .gitignore file

mivdittamo:

Add .class to the .gitignore to prevent .class files from being uploaded to the repo

### Is this fixing an issue?

none

none

- [  ] Test written/updating
- [  ] Tests passing

Just adding to .gitignore file, no additional logic added

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 target/
 .idea
+.class


.gitignore
```diff
@@ -1,2 +1,3 @@
 target/
 .idea
+.class
```
daniel-google, line 3: I'm not familiar with Maven, but normally any files produced as part of a build process should end up in their own directory (e.g. `/build), in which case that entire directory could be added to .gitignore.

Where exactly do the `.class` files end up when you build?

mivdittamo: I've been doing some testing with individual java files where I've been compiling and running them, and the .class files end up in the same directory as the the original java file. So far I've just been watching to make sure I don't commit those files, but I wanted to add this into the .gitignore to prevent them from accidentally getting uploaded to the repo.

daniel-google: Got it, thanks!

PR #6, opened 2020-06-30
Facebook Login Button with functionality and tests

jessdejong:

Added the Facebook login button on our app's homepage, and the Javascript code to make the button work. Additionally added the Jest javascript testing framework to the project, and some tests to test the facebook-login javascript code.

### Is this fixing an issue?

none

Added a couple of different config files to the project that are required for the Jest framework to work and build with our maven project. These files are supposed to be checked in to git.

- [  ] Test written/updating
- [  ] Tests passing

For future reference, tests were removed because Jest was running with Node, which didn't support the ES modules in the front-end. Additionally, Jest doesn't run/work with the Facebook SDK loading code, and we decided that it wasn't worth the effort to continue to debug the testing in order to make it work.

### Any additional comments?

Left some TODO comments, that I plan on filling in once the Datastore servlets get written in the backend.


src/main/webapp/FacebookAPI.js
```diff
@@ -0,0 +1,28 @@
+// Copyright 2019 Google LLC
```
daniel-google, line 1: nit: should these say 2020?

src/main/webapp/FacebookAPI.js
```diff
@@ -0,0 +1,28 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Handles calls to the Facebook API, provides wrapper methods with callbacks
+class FacebookAPI {
+  static getUserFacebookInfoAndRedirect(callback) {
+    return callback();
+  }
```
daniel-google, line 19: Hmm IIUC these aren't actually doing anything; it'll just always execute callback() immediately, so nothing is being mocked out -- it'll still call through to FB.

Instead I was thinking about something like:

```js
class FacebookApi {
  static getUserFacebookInfoAndRedirect(callback) {
    FB.api('/me?fields=id,name,email,friends', callback);
  }
}
```

Then use it like:

```js
FacebookApi.getUserFacebookInfoAndRedirect(function(response) {
  let userInfo = new URLSearchParams({
    name: response.name,
    id: response.id,
    email: response.email,
  });

  // Store user information in Datastore
  // TODO: Add post request
  // Redirect to the profile page, and pass the user's info in the URL
  window.location.href = "./profile.html?" + userInfo.toString();
});
```

And in your test, you can override it to invoke the callback with a mock response, something like:

```
FacebookApi.getUserFacebookInfoAndRedirect = function(callback) {
  callback({ /* construct a mock response here */ });
}; 
```

(my syntax may not be correct, I'm not really familiar with OO in Javascript)

jessdejong: The first two code snippets make sense, but the last is a little confusing. When testing this new function, what would I be inserting as the callback parameter? Also should the mock response be something similar to what I had before, like returning a string value?

src/main/webapp/facebook-login.js
```diff
@@ -0,0 +1,85 @@
...
+    statusChangeCallback(response);
+  });
+};
+
+// Async load of Facebook Javascript SDK
+// Provided by Facebook to load Javascript SDK
+(function(d, s, id) {
+  var js, fjs = d.getElementsByTagName(s)[0];
+  if (d.getElementById(id)) return;
+  js = d.createElement(s); js.id = id;
```
mivdittamo, line 48: nit: This could be on two separate lines (separated by the semicolon)

src/main/webapp/FacebookAPI.js
jessdejong: Also I've been a bit blocked on the fact that Jest doesn't 'work' (it errors immediately) when run on the code in `facebook-login.js` that loads the Facebook API async... so referencing the `FB` object in the `FacebookAPI.js` file doesn't pass any tests or compile the project. Not sure how to proceed here.

daniel-google: The last snippet is overriding the FacebookApi.getUserFacebookInfoAndRedirect function. But actually, maybe it'd be better to follow the Jest documentation -- it looks like there is first-class support for mocking:

https://jestjs.io/docs/en/mock-functions

(happy to help you figure it out if it's still unclear from those docs)

Also if your test code is failing to load the API, I suspect you'll want to separate the code which loads the API from the code which actually uses it -- so that you can avoid loading it in the test and just use the mock.

jessdejong:
@daniel-google @mivdittamo Addressed your commits, hopefully that works! I used the Jest mock functions for my tests.

src/main/webapp/test/facebook-login.test.js
```diff
@@ -0,0 +1,55 @@
...
+    // Create mock redirect URL
+    return "/profile?name=" + response.name + "&id=" + response.id + "&email=" + response.email;
+  });
+
+  let result = FacebookAPI.getUserFacebookInfoAndRedirect(mockCallback);
+  expect(result).toEqual("/profile?name=Tim&id=123&email=tim123@abcd.com");
+});
+
+// Test the function that responds to the logged in state of the user 
+test("Test checkLoginState function", () => {
```
daniel-google, line 39: These tests seem to only be exercising the test code itself, since you are mocking *both* the responses and the callbacks :P

The FacebookAPI mock methods look good, but you want to verify the *real* callback does what you're expecting. For example:

```js
test("Test checkLoginState function", () => {
  // Arrange: setup mocks for Facebook API methods.
  FacebookAPI.getUserFacebookInfoAndRedirect = jest.fn((callback) => {
    const response = {
      name: "Tim",
      id: "123",
      email: "tim123@abcd.com"
    };
    return callback(response);
  });
  FacebookAPI.checkLoginStatus = jest.fn((callback) => {
    const response = {
      login_status: "connected"
    };
    return callback(response);
  });

  // Act: Invoke the code-under-test.
  checkLoginState();

  // Assert: Verify the code-under-test had the expected output or side-effect(s).
  expect(window.location.href).toEqual(/* the expected redirect URL */);
});
```

(the [arrange/act/assert](http://wiki.c2.com/?ArrangeActAssert) comments aren't actually necessary, I'm just using them for clarity)

jessdejong:
@daniel-google Removed the testing, and separated the Facebook SDK load into a separate file. For future reference, tests were removed because Jest was running with Node, which didn't support the ES modules in the front-end. Additionally, Jest doesn't run/work with the Facebook SDK loading code, and we decided that it wasn't worth the effort to continue to debug the testing in order to make it work.

src/main/webapp/FacebookAPI.js
```diff
@@ -0,0 +1,38 @@
+// Copyright 2020 Google LLC
```
daniel-google, line 1: Might as well remove this now? And just use `FB` directly.

src/main/webapp/load-facebook.js
```diff
@@ -0,0 +1,42 @@
+// Copyright 2020 Google LLC
```
daniel-google, line 1: nit: can add this back to the main JS if you want

jessdejong: I'll just leave it in for now.

PR #7, opened 2020-06-30
Layout for Matches Page

alikew24:

{added a setup for the matches page that has a slideshow with placeholder images}

### Is this fixing an issue?

{none}

{js code will programatically add images later,this is just a layout for now}

- [  ] Test written/updating
- [  ] Tests passing

### Any additional comments?

{just html code, no tests necessary}

### DIFF
--- a/src/main/webapp/matches.html
+++ b/src/main/webapp/matches.html
@@ -3,11 +3,81 @@
   <head>
     <meta charset="UTF-8">
     <title>Friends of Friends</title>
+    <!-- Latest compiled and minified CSS -->
+    <link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/bootstrap/4.5.0/css/bootstrap.min.css">
+    <!-- jQuery library -->
+    <script src="https://ajax.googleapis.com/ajax/libs/jquery/3.5.1/jquery.min.js"></script>
+    <!-- Popper JS -->
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/popper.js/1.16.0/umd/popper.min.js"></script>
+    <!-- Latest compiled JavaScript -->
+    <script src="https://maxcdn.bootstrapcdn.com/bootstrap/4.5.0/js/bootstrap.min.js"></script>
     <link rel="stylesheet" href="style.css">
     <script src="script.js"></script>
   </head>
   <body>
-    <h1>Matches</h1>
+    <div class="container-fluid">
+      <nav class="navbar navbar-expand-sm justify-content-center navbar-background">
+        <ul class="navbar-nav">
+          <li class="nav-item">
+            <img src="images/logo.jpg" class="rounded-circle" alt="logo" height="80" width="100">
+          </li>
+          <li class="nav-item">
+            <a class="nav-link nav-text-style" href="profile.html">Profile</a>
+          </li>
+          <li class="nav-item">
+            <a class="nav-link nav-text-style" href="feed.html">Feed</a>
+          </li>
+          <li class="nav-item">
+            <a class="nav-link nav-text-style" href="matches.html">Matches</a>
+          </li>
+        </ul>
+      </nav>
+      <div class="content">
+        <h1>Matches</h1>
+        <div class="row row-cols-5">
+          <div class="col card card-container m-5">
+            <img class="card-img-top" src="images/placeholder.jpg" alt="Card image">
+            <div class="card-body">
+              <h4 class="card-title">Match 1 Name</h4>
+              <p class="card-text">Insert match 1 bio here</p>
+              <a href="#" class="btn see-profile-btn">See Profile</a>
+            </div>
+          </div>
+          <div class="col card card-container-two m-5">
+            <img class="card-img-top" src="images/placeholder.jpg" alt="Card image">
+            <div class="card-body">
+              <h4 class="card-title">Match 1 Name</h4>
+              <p class="card-text">Insert match 1 bio here</p>
+              <a href="#" class="btn see-profile-btn-two">See Profile</a>
+            </div>
+          </div>
+          <div class="col card card-container m-5">
+            <img class="card-img-top" src="images/placeholder.jpg" alt="Card image">
+            <div class="card-body">
+              <h4 class="card-title">Match 1 Name</h4>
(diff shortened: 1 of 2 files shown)


alikew24:
typo up there ^ there is no slideshow, but there are cards with placeholder images that will later be programatically implemented

src/main/webapp/style.css
```diff
@@ -40,3 +40,18 @@ h1 {
     align-self: center;
     margin-left: 625px;
 }
+.card-container {
+    /* width: 30%; */
+    justify-content: center;
+    background-color: #fbe8a6;
```
jessdejong, line 46: nit: alphabetize css properties

src/main/webapp/style.css
```diff
@@ -40,3 +40,18 @@ h1 {
...
+}
+.see-profile-btn {
+    background-color: #b4dfe5;
+}
+.card-container-two {
+    justify-content: center;
+    background-color: #b4dfe5;
```
jessdejong, line 53: nit: same here

src/main/webapp/style.css
```diff
@@ -40,3 +40,18 @@ h1 {
...
+}
+.see-profile-btn-two {
+    background-color: #fbe8a6;
+}
```
jessdejong, line 57: nit: add new line

alikew24:
haha ya i realized that after I requested a review. Fixed those nits!

src/main/webapp/matches.html
```diff
@@ -3,11 +3,81 @@
   <head>
     <meta charset="UTF-8">
     <title>Friends of Friends</title>
+    <!-- Latest compiled and minified CSS -->
```
daniel-google, line 6: nit: this URL isn't actually pointing to the "latest", but a specific version -- so remove that from the comment

(alternatively, remove all of these comments since it's clear from the URL which library they're providing)

src/main/webapp/matches.html
```diff
@@ -3,11 +3,81 @@
...
+    <link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/bootstrap/4.5.0/css/bootstrap.min.css">
+    <!-- jQuery library -->
+    <script src="https://ajax.googleapis.com/ajax/libs/jquery/3.5.1/jquery.min.js"></script>
+    <!-- Popper JS -->
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/popper.js/1.16.0/umd/popper.min.js"></script>
+    <!-- Latest compiled JavaScript -->
```
daniel-google, line 12: nit: here too

src/main/webapp/matches.html
```diff
@@ -3,11 +3,81 @@
...
+    <script src="https://maxcdn.bootstrapcdn.com/bootstrap/4.5.0/js/bootstrap.min.js"></script>
     <link rel="stylesheet" href="style.css">
     <script src="script.js"></script>
   </head>
   <body>
-    <h1>Matches</h1>
+    <div class="container-fluid">
+      <nav class="navbar navbar-expand-sm justify-content-center navbar-background">
+        <ul class="navbar-nav">
+          <li class="nav-item">
```
daniel-google, line 21: nit: rather than classing these all with `nav-item`, maybe apply the styling to `.navbar-nav li` (and similarly you could replace `.navbar-nav` in favor of `.nav ul`)

up to you though, this is just a matter of preference

src/main/webapp/matches.html
```diff
@@ -3,11 +3,81 @@
...
+              <p class="card-text">Insert match 1 bio here</p>
+              <a href="#" class="btn see-profile-btn">See Profile</a>
+            </div>
+          </div>
+          <div class="col card card-container-two m-5">
+            <img class="card-img-top" src="images/placeholder.jpg" alt="Card image">
+            <div class="card-body">
+              <h4 class="card-title">Match 1 Name</h4>
+              <p class="card-text">Insert match 1 bio here</p>
+              <a href="#" class="btn see-profile-btn-two">See Profile</a>
```
daniel-google, line 51: nit: instead of having two classes to apply alternating styles (`see-profile-btn` and `see-profile-btn-two`), you can use even-odd rules: https://www.w3.org/Style/Examples/007/evenodd.en.html

src/main/webapp/matches.html
```diff
@@ -3,11 +3,81 @@
...
+            <div class="card-body">
+              <h4 class="card-title">Match 1 Name</h4>
+              <p class="card-text">Insert match 1 bio here</p>
+              <a href="#" class="btn see-profile-btn">See Profile</a>
+            </div>
+          </div>
+          <div class="col card card-container-two m-5">
+            <img class="card-img-top" src="images/placeholder.jpg" alt="Card image">
+            <div class="card-body">
+              <h4 class="card-title">Match 1 Name</h4>
```
daniel-google, line 49: Match 2 Name?

(similarly below)

src/main/webapp/matches.html
```diff
@@ -3,11 +3,81 @@
...
+          </li>
+          <li class="nav-item">
+            <a class="nav-link nav-text-style" href="matches.html">Matches</a>
+          </li>
+        </ul>
+      </nav>
+      <div class="content">
+        <h1>Matches</h1>
+        <div class="row row-cols-5">
+          <div class="col card card-container m-5">
```
daniel-google, line 38: Are these placeholders going to be populated in JS later? Assuming yes, you might as well construct them with JS now and just replace the data source later?

(it's probably better to support a dynamic number rather than assuming there are 5, anyway)

src/main/webapp/matches.html
alikew24: nav-item is a class from bootstrap that I am utilizing! same with navbar-nav

daniel-google: ah makes sense!

src/main/webapp/matches.html
alikew24: yep- the 5 cards right now are just to see and visualize what the page will look like. For now I am focusing on finishing the profile page and slideshow page, and then after the mvp focus on matches page. This is just to kinda visualize what it should look like after the functionality is implemented.

alikew24:
added these changes!

PR #8, opened 2020-06-30
Feed Page Layout

alikew24:

Created a layout for the feed page with a template slideshow and placeholder images!

### Is this fixing an issue?

nope

The slideshow images will be programatically implemented later with javascript, for now this is just a template

- [  ] Test written/updating
- [  ] Tests passing
- [  x] Coding style (indentation, etc)


src/main/webapp/style.css
```diff
@@ -40,3 +40,12 @@ h1 {
...
 }
+.friend-me-btn {
+    background-color: #b4dfe5;
+}
+.pass-btn {
+    background-color: #fbe8a6;
+}
+.button-container {
+    justify-content: center;
+}
```
jessdejong, line 51: nit: add newline at end of file

src/main/webapp/feed.html
```diff
@@ -3,11 +3,83 @@
...
+              </div> 
+            </div>
+          </div>
+          <a class="carousel-control-prev" href="#feed" data-slide="prev">
+            <span class="carousel-control-prev-icon"></span>
+          </a>
+          <a class="carousel-control-next" href="#feed" data-slide="next">
+            <span class="carousel-control-next-icon"></span>
+          </a>
+        </div> <!--carousel -->
```
jessdejong, line 72: nit: should this comment be on it's own line?

alikew24: the comment is for the closing div so that anyone looking at the code knows what the div is closing!

alikew24:
added that nit!

src/main/webapp/feed.html
```diff
@@ -3,11 +3,83 @@
...
     <meta charset="UTF-8">
     <title>Friends of Friends</title>
+     <!-- Latest compiled and minified CSS -->
+    <link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/bootstrap/4.5.0/css/bootstrap.min.css">
+    <!-- jQuery library -->
+    <script src="https://ajax.googleapis.com/ajax/libs/jquery/3.5.1/jquery.min.js"></script>
+    <!-- Popper JS -->
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/popper.js/1.16.0/umd/popper.min.js"></script>
+    <!-- Latest compiled JavaScript -->
+    <script src="https://maxcdn.bootstrapcdn.com/bootstrap/4.5.0/js/bootstrap.min.js"></script>
```
daniel-google, line 13: nit: same as last CL, remove "latest" or comments altogether

Can you think of a good way to share common elements like this across pages without duplicating code?

src/main/webapp/feed.html
```diff
@@ -3,11 +3,83 @@
...
     <link rel="stylesheet" href="style.css">
     <script src="script.js"></script>
   </head>
   <body>
-    <h1>Feed</h1>
+    <div class="container-fluid">
+      <nav class="navbar navbar-expand-sm justify-content-center navbar-background">
```
daniel-google, line 19: nit: same comments as other CL re: classes vs type qualifiers, and same comment as above regarding code duplication

src/main/webapp/feed.html
```diff
@@ -3,11 +3,83 @@
...
+                  <p>This is an example bio.</p>
+                </div> 
+            </div>
+            <div class="carousel-item">
+              <img src="images/placeholder.jpg" alt="placeholder" width="1100" height="800">
+              <div class="carousel-caption">
+                <h3>NAME</h3>
+                <p>This is an example bio.</p>
+              </div> 
+            </div>
```
daniel-google, line 64: Same comment as other CL re: generating these in Javascript -- it's probably better than hard-coding them, since you can just swap out the data source later instead of checking-in this HTML and later changing it to Javascript.

src/main/webapp/feed.html
alikew24: these are also all from bootstrap!

src/main/webapp/feed.html
alikew24: I generated these in javascript in a different branch that I havent submitted a pr for yet because I still have to test it- for this PR I just wanted to set up a general idea of what the page would look like !

src/main/webapp/feed.html
daniel-google: I actually brought this up with Nik today, we found a JS library that looks useful:

https://github.com/LexmarkWeb/csi.js

Perhaps worth trying out for the common layout elements (e.g. navigation) in a follow-up PR; you can probably just leave `<script>` and `<link>` tags as-is though.

alikew24: ok! I will add this in a follow up once these get merged in

alikew24:
deleted the comments

PR #9, opened 2020-07-01
Create map of friendships for the initial potential match finding process

mivdittamo:

- Create a map of all the direct Facebook friendships found

- Create object to represent a single user and their set of Facebook friends

- Add Guava dependency to project

### Is this fixing an issue?

none

- The map created here is to be used later for the initial potential match (2nd friends) finding process.

- [ X ] Test written/updating
- [ X ] Tests passing
- [ X ] Coding style (indentation, etc)


src/main/java/com/google/sps/data/friend_graph/UserFriendsMap.java
```diff
@@ -20,45 +20,48 @@
...
 
   /**
-  * Constructor for the UserFriendsGraph class
+  * Constructor for the UserFriendsMap class
   *
-  * Forms the graph which is represented by a map of UserIDs to a set
-  * ID's of that user's friends.
+  * Forms a map of UserIDs to a set ID's of that user's friends.
   * 
   * Map created in this class will be used to find a user's set of potential
```
alikew24: nit: might be useful to explain the what the keys represent in this map

src/main/java/com/google/sps/data/friend_graph/UserNode.java
```diff
@@ -59,15 +51,16 @@ public boolean equals(Object obj) {
...
+      this.currentUserFriendIDs.equals(otherUser.currentUserFriendIDs);
   }
 
   @Override
   public int hashCode() {
     int hashTotal = 0;
-    for (int i = 0; i < this.userID.length(); i++) {
-      hashTotal += (i * (int)this.userID.charAt(i));
-    }
+    final int prime = 31;
```
alikew24: nit: explain where the number 31 comes from in a comment

mivdittamo: Added a comment on this line that explains it! Main idea here was using a prime number to help calculate a more unique hash value.

src/main/java/com/google/sps/data/friend_map/UserNode.java
```diff
@@ -0,0 +1,66 @@
...
+  *
+  * @param userID The unique id number of each user (obtained from Facebook API)
+  * @param currentUserFriendIDs Immutable set of all of the IDs of a user's friends
+  */
+  public UserNode(String userID, Collection<String> currentUserFriendIDs) {
+    this.userID = userID;
+    this.currentUserFriendIDs = ImmutableSet.copyOf(currentUserFriendIDs);
+  }
+
+  public String getUserID() {
```
estep-google, line 39: I would just say `getId()` - we know from the class name that this is a user.

src/main/java/com/google/sps/data/friend_map/UserNode.java
```diff
@@ -0,0 +1,66 @@
...
+    return this.userID;
+  }
+
+  public ImmutableSet<String> getCurrentUserFriendIDs() {
```
estep-google, line 43: This class is essentially immutable (you have no setters). Do we need to say current? We certainly don't need user here. To me, this can just be `getFriendsIds()`

src/main/java/com/google/sps/data/friend_map/UserNode.java
```diff
@@ -0,0 +1,66 @@
...
+  }
+
+  @Override
+  public boolean equals(Object obj) {
+    if (!(obj instanceof UserNode)) {
+      return false;
+    }
+    UserNode otherUser = (UserNode) obj;
+    return this.userID.equals(otherUser.userID) &&
+      this.currentUserFriendIDs.equals(otherUser.currentUserFriendIDs);
```
estep-google, line 54: To check: Do you know what the `equals` in the `ImmutableSet` class does?

src/main/java/com/google/sps/data/friend_map/UserNode.java
```diff
@@ -0,0 +1,66 @@
...
+  }
+
+  @Override
+  public int hashCode() {
+    int hashTotal = 0;
+    final int prime = 31; // use of prime number to create more unique hash values
+    hashTotal += prime * (this.userID == null ? 0 : this.userID.hashCode());
```
estep-google, line 61: Why would `userID` be null?

src/main/java/com/google/sps/data/friend_map/UserNode.java
```diff
@@ -0,0 +1,66 @@
...
+    hashTotal += prime * (this.currentUserFriendIDs == null ? 0 : this.currentUserFriendIDs.hashCode());
```
estep-google, line 62: Why would `currentUserFriendsIDs` be null?

src/main/java/com/google/sps/data/friend_map/UserFriendsMap.java
```diff
@@ -0,0 +1,68 @@
...
+  private Map<String, ImmutableSet<String>> friendMap;
+
+  /**
+  * Constructor for the UserFriendsMap class
+  *
+  * Forms a map of UserIDs to a set ID's of that user's friends.
+  *
+  * Map keys are the UserIDs which are unique identifiers for each user (reprsented
+  * as a string here) and the map values are the set of IDs of each friend of the 
+  * current user (represented here as an ImmutableSet of Strings).
```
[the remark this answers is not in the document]
mivdittamo: got it!

src/main/java/com/google/sps/data/friend_map/UserFriendsMap.java
```diff
@@ -0,0 +1,68 @@
...
+  * 
+  * Map created in this class will be used to find a user's set of potential
```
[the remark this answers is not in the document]
mivdittamo: When I move this to be a class level comment, should I add the <p> there as well?

src/main/java/com/google/sps/data/friend_map/UserFriendsMap.java
```diff
@@ -0,0 +1,68 @@
...
+
+    for (UserNode user: userNodes) {
+      String userID = user.getUserID();
+      ImmutableSet<String> userFriendIDs = user.getCurrentUserFriendIDs();
+      friendMap.put(userID, userFriendIDs);
+    }
+  }
+
+  public Map<String, ImmutableSet<String>> getFriendMap() {
+    return this.friendMap;
```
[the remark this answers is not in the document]
mivdittamo: Got it! I'll remove `this` from the getter methods for this class.

src/main/java/com/google/sps/data/friend_map/UserNode.java
```diff
@@ -0,0 +1,66 @@
...
+/**
+* Class to represent a User in the map representing all user friendships
+*/
+public class UserNode {
+
+  private final String userID;
+  private ImmutableSet<String> currentUserFriendIDs;
+
+  /**
+  * Constructs an instance of a UserNode
```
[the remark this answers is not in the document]
mivdittamo: I looked up common practices for javadocs for constructors and most seem to have a description before they list the parameters. I can change this line though to say 'Creates a node to represent a user with their set of friends'

src/main/java/com/google/sps/data/friend_map/UserNode.java
```diff
@@ -0,0 +1,66 @@
...
+import java.util.Collection;
+import com.google.common.collect.ImmutableSet;
+
+/**
+* Class to represent a User in the map representing all user friendships
+*/
+public class UserNode {
+
+  private final String userID;
+  private ImmutableSet<String> currentUserFriendIDs;
```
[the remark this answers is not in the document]
mivdittamo: Got it!

src/main/java/com/google/sps/data/friend_map/UserNode.java
mivdittamo: Got it!

src/main/java/com/google/sps/data/friend_map/UserNode.java
mivdittamo: Got it! I'll also go ahead and change the field name to friendIDs

src/main/java/com/google/sps/data/friend_map/UserNode.java
mivdittamo: I checked it out and it looks like ImmutableSet equals method would end up calling the Sets.equalsImp method (because the isHashCodeFast() method looks like it will go ahead ans return false?), and that will check if the sizes are the two sets are equal and if they contain all the same elements.
Should I still be using the ImmutableSets equals method here or is there a more efficient way to check these two sets?

src/main/java/com/google/sps/data/friend_map/UserNode.java
mivdittamo: When I looked up how to write a good hashCode method and checking for nulls like this was something I saw in the example I found. We wouldn't be passing in null values to the UserNode constructor, but I kept the null checks just in case null values came up in error.

Would it be better though to have a null check in the UserNode constructor and then the null checks in the hashCode method can be removed?

src/main/java/com/google/sps/data/friend_map/UserNode.java
mivdittamo: Same as above

src/main/java/com/google/sps/data/friend_map/UserNode.java
estep-google: The "two sets are equal if they contain all the same elements" is where my concern lies. How does it determine that? I assume it's probably only doing memory address comparisons, which means each object will only be considered equal to itself, but not equal to a different object that contains the exact same data.

src/main/java/com/google/sps/data/friend_map/UserNode.java
estep-google: I think it is a reasonable assumption that the inputs to the constructor here are not null. I would annotate each with `@Nonnull` (`javax.annotations.Nonnull`). Then I would still `checkNotNull` (https://guava.dev/releases/19.0/api/docs/com/google/common/base/Preconditions.html#checkNotNull(T)). This is all in the constructor.

You can statically import `checkNotNull`.

src/main/java/com/google/sps/data/friend_map/UserNode.java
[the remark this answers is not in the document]
estep-google: Yeah, better.

src/main/java/com/google/sps/data/friend_map/UserFriendsMap.java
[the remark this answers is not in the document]
estep-google: Yep! `<p>` is standard practice no matter where the javadoc comment is

src/main/java/com/google/sps/data/friend_map/UserNode.java
mivdittamo: It's using a containsAll method, and from my understanding, contains and containsAll methods use the equals method for the objects they're comparing. So I'd assume based on this that the String class's equals method would be used to compare the elements within these sets.

src/main/java/com/google/sps/data/friend_map/UserFriendsMap.java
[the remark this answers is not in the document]
mivdittamo: Got it! Thanks!

src/main/java/com/google/sps/data/friend_map/UserFriendsMap.java
```diff
@@ -0,0 +1,68 @@
...
+  * current user (represented here as an ImmutableSet of Strings).
+  * 
+  * Map created in this class will be used to find a user's set of potential
+  * matches (2nd friends-> those who are friends with a user's current friend
+  * who is not already friends with the user).
+  *
+  * @param userNodes A set of UserNodes that are used to form the graph structure
+  */
+  public UserFriendsMap (Set<UserNode> userNodes) {
+    this.friendMap = new HashMap<>();
```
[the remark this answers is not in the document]
mivdittamo: Got it! I ended up using the Java streams way for this constructor.

src/main/java/com/google/sps/data/friend_map/UserNode.java
mivdittamo: Got it! Added the annotations and used the checkNotNull method.

src/main/java/com/google/sps/data/friend_map/UserFriendsMap.java
```diff
@@ -20,49 +20,41 @@
 import java.util.Map;
 import java.util.HashMap;
 import com.google.common.collect.ImmutableSet;
+import com.google.common.collect.ImmutableMap;
 
-/*
-* A map of all the direct friendships between users
+/**
+* <p>A map of all the direct friendships between users which will be used
```
estep-google, line 26: This is only for subsequent paragraphs. The first paragraph doesn't need the `<p>`

src/main/java/com/google/sps/data/friend_map/UserFriendsMap.java
```diff
@@ -20,49 +20,41 @@
...
-  * Map keys are the UserIDs which are unique identifiers for each user (reprsented
-  * as a string here) and the map values are the set of IDs of each friend of the 
-  * current user (represented here as an ImmutableSet of Strings).
-  * 
-  * Map created in this class will be used to find a user's set of potential
-  * matches (2nd friends-> those who are friends with a user's current friend
-  * who is not already friends with the user).
-  *
-  * @param userNodes A set of UserNodes that are used to form the graph structure
+  * @param userNodes A set of {@code UserNodes} that are used to form the map
```
estep-google, line 40: You don't need a `{@code}` reference in the param. Javadoc generation will take care of that.

src/main/java/com/google/sps/data/friend_map/UserNode.java
```diff
@@ -15,33 +15,36 @@
...
 */
 public class UserNode {
 
   private final String userID;
-  private ImmutableSet<String> currentUserFriendIDs;
+  private final ImmutableSet<String> friendIDs;
 
   /**
-  * Constructs an instance of a UserNode
+  * Creates a node to represent a user with their set of friends
```
estep-google, line 32: Verbage - drop the "c" ("Create a node...")

src/test/java/com/google/sps/data/friend_map/FriendsMapCreationTest.java
```diff
@@ -29,6 +29,7 @@
 import org.junit.runner.RunWith;
 import org.junit.runners.JUnit4;
 import com.google.common.collect.ImmutableSet;
+import com.google.common.collect.ImmutableMap;
```
estep-google, line 32: I'm having trouble figuring out how to add a comment in the right place - apparently I shouldn't have put this off :)

Your set up method isn't usually how this works. Having some statics at the top to use is a good idea. Normally you then construct the specifics for a test case within the test case itself. This keeps the scenario right in front of the reader instead of having to look back up to somewhere else. This works with good static variable names so it's clear what they are too.

src/main/java/com/google/sps/data/friend_map/UserFriendsMap.java
mivdittamo: Got it!

src/main/java/com/google/sps/data/friend_map/UserFriendsMap.java
mivdittamo: Got it!

src/main/java/com/google/sps/data/friend_map/UserNode.java
mivdittamo: Got it!

src/test/java/com/google/sps/data/friend_map/FriendsMapCreationTest.java
mivdittamo: Ok! Since most of what is currently in the setUp method could just be constructed within the test cases that they apply to, do I still need to keep the setUp method here?

estep-google: No, you likely don't need it at all for a test class like this.

mivdittamo: Ok, got it!

src/test/java/com/google/sps/data/friend_map/FriendsMapCreationTest.java
```diff
@@ -48,85 +45,75 @@
...
+
     Set<UserNode> oneUserSet = new HashSet<>();
     oneUserSet.add(userA);
     UserFriendsMap oneUserFriendsMap = new UserFriendsMap(oneUserSet);
 
     Map<String, ImmutableSet<String>> expectedMap = new HashMap<>();
-    expectedMap.put(USER_A_ID, emptyStringSet);
+    expectedMap.put(USER_A_ID, ImmutableSet.of());
 
     Assert.assertEquals(oneUserFriendsMap.getFriendMap(), ImmutableMap.copyOf(expectedMap));
```
estep-google, line 66: Can you look up if you have a way to use `assertThat`? It allows for a more "natural reading" of test cases. Example:

`assertThat(oneUserFriendsMap.getFriendMap()).containsExactly(expectedMap);`

src/test/java/com/google/sps/data/friend_map/FriendsMapCreationTest.java
```diff
@@ -48,85 +45,75 @@
...
   */
   @Test
   public void oneUserGraphTest() {
+    userA = new UserNode(USER_A_ID, ImmutableSet.of());
+
     Set<UserNode> oneUserSet = new HashSet<>();
     oneUserSet.add(userA);
     UserFriendsMap oneUserFriendsMap = new UserFriendsMap(oneUserSet);
 
     Map<String, ImmutableSet<String>> expectedMap = new HashMap<>();
```
estep-google, line 63:
```
ImmutableMap<String, ImmutableSet<String>> expectedMap =
    ImmutableMap.of(
        USER_A_ID, ImmutableSet.of()
    );
```

src/test/java/com/google/sps/data/friend_map/FriendsMapCreationTest.java
```diff
@@ -48,85 +45,75 @@
...
   }
 
   /**
-  * Tests if graph is forming correctly when there are just 3 users, 2 of which are
+  * Tests if map is forming correctly when there are just 3 users, 2 of which are
   * friends with the third user. There are no connections other than this.
   *
-  * Should result in a UserFriendsGraph with three keys:
+  * <p>Should result in a UserFriendsMap with three keys:
   *    User C should map to a friend set with just User E
```
estep-google, line 97: Up to you, but there's no reason you can't use users A, B, C here. Reusing is totally fine.

src/test/java/com/google/sps/data/friend_map/FriendsMapCreationTest.java
```diff
@@ -48,85 +45,75 @@
   private UserNode userC;
```
estep-google, line 45: These should not be here. You want to declare in each test case.

Instead of `userA = new UserNode(...);`, you'd want `UserNode userA = new UserNode(...);`

src/test/java/com/google/sps/data/friend_map/FriendsMapCreationTest.java
mivdittamo: Got it! I went ahead and did this for the expected maps for the other two tests as well.

src/test/java/com/google/sps/data/friend_map/FriendsMapCreationTest.java
mivdittamo: Ok, got it!

src/test/java/com/google/sps/data/friend_map/FriendsMapCreationTest.java
mivdittamo: Ok, got it! I took these UserNode declarations out and moved them to be declared within the test cases.

src/test/java/com/google/sps/data/friend_map/FriendsMapCreationTest.java
mivdittamo: I found this resource for how to use assertThat for maps: https://mkyong.com/unittest/junit-how-to-test-a-map/
I'm adding this in now!

pom.xml
```diff
@@ -38,12 +38,24 @@
       <artifactId>junit</artifactId>
       <version>4.12</version>
       <scope>test</scope>
+      <exclusions>
+        <exclusion>
+          <groupId>org.hamcrest</groupId>
```
estep-google, line 43: Sorry, I'm going to be picky here :)

Let's use Truth instead of Hamcrest (https://truth.dev/) (I forgot about Hamcrest or I would have been more clear earlier)

src/test/java/com/google/sps/data/friend_map/FriendsMapCreationTest.java
```diff
@@ -30,21 +30,16 @@
 import org.junit.runners.JUnit4;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.ImmutableMap;
+import static org.hamcrest.CoreMatchers.is;
```
estep-google, line 33: Static imports go above. Ordering is:

```
package com.awesome.project....

import static ...
import static ...

import ...
import ...
```

src/test/java/com/google/sps/data/friend_map/FriendsMapCreationTest.java
```diff
@@ -74,48 +71,52 @@ public void oneUserGraphTest() {
   */
   @Test
   public void twoUsersNoConnectionsGraphTest() {
-    userA = new UserNode(USER_A_ID, ImmutableSet.of());
-    userB = new UserNode(USER_B_ID, ImmutableSet.of());
+    UserNode userA = new UserNode(USER_A_ID, ImmutableSet.of());
+    UserNode userB = new UserNode(USER_B_ID, ImmutableSet.of());
 
     Set<UserNode> twoUserSet = new HashSet<>();
```
estep-google, line 77: If you want, this can be a little more compact:

```
Set<UserNode> twoUserSet = new HashSet<>(Arrarys.asList(userA, userB));
```

You could also use my favorite word - immutable! Your constructor doesn't modify the set, so you can safely do this because `ImmutableSet` implements the `Set` interface. For reference, `ImmutableSet` throws an `UnsupportedOperationException` on any method that would otherwise mutate the set (e.g., `add()`).

```
Set<UserNode> twoUserSet = ImmutableSet.of(userA, userB);
```

src/test/java/com/google/sps/data/friend_map/FriendsMapCreationTest.java
```diff
@@ -74,48 +71,52 @@ public void oneUserGraphTest() {
...
 
   /**
   * Tests if map is forming correctly when there are just 3 users, 2 of which are
   * friends with the third user. There are no connections other than this.
   *
   * <p>Should result in a UserFriendsMap with three keys:
-  *    User C should map to a friend set with just User E
-  *    User D should map to a friend set with just User E
-  *    User E should map to a friend set with User C and User D.
+  *    User A should map to a friend set with just User C
```
estep-google, line 96: This should be an unordered list.

src/test/java/com/google/sps/data/friend_map/FriendsMapCreationTest.java
```diff
@@ -74,48 +71,52 @@ public void oneUserGraphTest() {
...
   @Test
   public void threeUsersTwoConnectionGraphTest() {
-    userC = new UserNode(USER_C_ID, ImmutableSet.of(USER_E_ID));
-    userD = new UserNode(USER_D_ID, ImmutableSet.of(USER_E_ID));
-    userE = new UserNode(USER_E_ID, ImmutableSet.of(USER_C_ID, USER_D_ID));
+    UserNode userA = new UserNode(USER_A_ID, ImmutableSet.of(USER_C_ID));
+    UserNode userB = new UserNode(USER_B_ID, ImmutableSet.of(USER_C_ID));
+    UserNode userC = new UserNode(USER_C_ID, ImmutableSet.of(USER_A_ID, USER_B_ID));
     
     Set<UserNode> threeUsersTwoConnectionSet = new HashSet<>();
```
estep-google, line 106: See earlier comment about set construction.

pom.xml
mivdittamo: Ok, switched to using Truth instead. I used the containsExactlyEntriesIn method with assertThat.

src/test/java/com/google/sps/data/friend_map/FriendsMapCreationTest.java
mivdittamo: Got it!

src/test/java/com/google/sps/data/friend_map/FriendsMapCreationTest.java
mivdittamo: Got it! I changed this to form an ImmutableSet.

src/test/java/com/google/sps/data/friend_map/FriendsMapCreationTest.java
mivdittamo: Just to clarify, should the javadocs be changed to say that 'User A should map to an unordered list with just User C'? The values in the map are sets.

src/test/java/com/google/sps/data/friend_map/FriendsMapCreationTest.java
mivdittamo: Got it!

src/test/java/com/google/sps/data/friend_map/FriendsMapCreationTest.java
mivdittamo: Got it!

pom.xml
```diff
@@ -38,22 +38,16 @@
       <artifactId>junit</artifactId>
       <version>4.12</version>
       <scope>test</scope>
-      <exclusions>
```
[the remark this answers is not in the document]
mivdittamo: There wasn't anything in https://truth.dev/ about adding an exclusion, just adding a dependency. The original exclusion I had was for something with Hamcrest.

PR #10, opened 2020-07-06
Create Nav Bar programatically

alikew24:

created a nav bar html file and loaded it onto the profile page with a script command

### Is this fixing an issue?

none

none

- [  ] Test written/updating
- [  ] Tests passing
- [  x] Coding style (indentation, etc)


src/main/webapp/nav.html
```diff
@@ -0,0 +1,26 @@
...
+            <a class="nav-link nav-text-style" href="profile.html">Profile</a>
+          </li>
+          <li class="nav-item">
+            <a class="nav-link nav-text-style" href="feed.html">Feed</a>
+          </li>
+          <li class="nav-item">
+            <a class="nav-link nav-text-style" href="matches.html">Matches</a>
+          </li>
+        </ul>
+</nav>
```
mivdittamo, line 25: nit: adjust indentation

src/main/webapp/style.css
```diff
@@ -36,7 +36,5 @@ h1 {
     width: 30%;
 }
 .image-upload {
-    align-content: center;
-    align-self: center;
-    margin-left: 625px;
+    justify-content: center;
```
mivdittamo, line 39: nit: alphabetize css properties

alikew24: i deleted the other css properties, so the justify-content center is actually the only property in that block!

mivdittamo: Oh whoops! Sorry about that! You're good here!

alikew24:
fixed the indentation!

src/main/webapp/nav.html
```diff
@@ -0,0 +1,27 @@
+<!DOCTYPE html>
```
daniel-google, line 1: Since this is being included directly into the other pages, I think you'll want to remove all tags except `<nav>` and its contents (i.e. this is not a stand-alone page itself, so it doesn't need to be "valid" HTML on its own).

Otherwise the resulting page will have multiple nested `<html>`, `<body>`, etc. (unless the CSI library somehow removes this?).

alikew24:
Removed the tags!

PR #12, opened 2020-07-07
Current User Function

alikew24:

added a current user function that takes in the url query parameter to get the id of the current user, and puts that id into the other links

### Is this fixing an issue?

none

none

- [  ] Test written/updating
- [  ] Tests passing


src/main/webapp/script.js
```diff
@@ -12,3 +12,25 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+function getCurrentUser() {
+  const params = new URLSearchParams(window.location.search);
+  uid = params.get('id');
+  console.log(uid);
```
jessdejong, line 18: nit: remove console.log()?

src/main/webapp/script.js
```diff
@@ -12,3 +12,25 @@
...
+  return uid;
+}
+function setProfileLink() {
+  let uid = getCurrentUser();
+  var link = document.getElementById("profileLink");
+  link.setAttribute('href', 'profile.html?id=' + uid);
```
jessdejong, line 24: Maybe: `link.setAttribute('href', 'profile.html?id=' + getCurrentUser);`

src/main/webapp/script.js
```diff
@@ -12,3 +12,25 @@
...
+}
+function setFeedLink() {
+  let uid = getCurrentUser();
+  var link = document.getElementById("feedLink");
+  link.setAttribute('href', 'feed.html?id=' + uid);
+}
```
jessdejong, line 30: nit: new line between functions?
[Google javascript style functions](https://google.github.io/styleguide/jsguide.html#formatting-function-arguments)

src/main/webapp/script.js
alikew24: i know my way is more code but i like the way its organized better!

src/main/webapp/nav.html
```diff
@@ -4,13 +4,13 @@
       <img src="images/logo.jpg" class="rounded-circle" alt="logo" height="80" width="100">
     </li>
     <li class="nav-item">
-      <a class="nav-link nav-text-style" href="profile.html">Profile</a>
+      <a class="nav-link nav-text-style" href="#" onclick="setProfileLink()" id="profileLink">Profile</a>
```
daniel-google, line 7: Instead of changing the `href` in `#onClick`, which will break some behavior (e.g. Right-click -> Open in new tab or Copy Link Address), it'd be better to do it when the page first loads.

src/main/webapp/script.js
```diff
@@ -12,3 +12,27 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+function getCurrentUser() {
+  const params = new URLSearchParams(window.location.search);
+  uid = params.get('id');
+  return uid;
+}
+
+function setProfileLink() {
```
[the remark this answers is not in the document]
daniel-google: I see this was marked as resolved but nothing was changed?

src/main/webapp/nav.html
daniel-google: Followed-up offline, tried out MutationObserver but couldn't get it to work. Fine with punting on this to avoid blocking you!

PR #13, opened 2020-07-09
Storing User Information using Datastore

jessdejong:

Added the UserDataServlet.java servlet to handle updating/creating User entities to store in Datastore, as well as retrieve user information of user's that already are placed in Datastore. In the front-end, added post requests to the servlet as soon as the user logs in, as well as when they update their profile on the profile page. Lastly, added jUnit tests to test UserDataServlet.

### Is this fixing an issue?

none

* The fetch post request after facebook login sends data via the query URL. After the MVP, I will consider changing this method to possibly use jQuery and Ajax to send JSON to the server, especially since a user's friend-list needs to be sent in this information.
* https://cloud.google.com/appengine/docs/standard/java/tools/localunittesting documents how to use `LocalServiceTestHelper` in order to test Datastore using just local memory. I used this to test Datastore in my tests.

- [ x ] Test written/updating
- [ x ] Tests passing

### Any additional comments?


src/main/java/com/google/sps/servlets/UserDataServlet.java
```diff
@@ -15,18 +16,142 @@
...
+import com.google.appengine.api.datastore.DatastoreService;
+import com.google.appengine.api.datastore.DatastoreServiceFactory;
+import com.google.appengine.api.datastore.Entity;
+import com.google.appengine.api.datastore.PreparedQuery;
+import com.google.appengine.api.datastore.Query;
+import com.google.gson.Gson;
 
-/** Servlet to add user profile data to the Datastore */
-@WebServlet("/data")
+/** Servlet that provides information about a specific user, and allows setting a user's info */
```
alikew24, line 34: nit: add details about what info i.e name, bio, id, photos

src/main/java/com/google/sps/servlets/UserDataServlet.java
```diff
@@ -15,18 +16,142 @@
...
+    response.setContentType("application/json");
+    response.getWriter().println(gson.toJson(userData));
+  }
+
+  @Override
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    // Get the user's info to either update or create
+    String userId = getStringParameter(request, USER_ID_PROPERTY, DEFAULT_STRING);
+    String userName = getStringParameter(request, USER_NAME_PROPERTY, DEFAULT_STRING);
+    String userEmail = getStringParameter(request, USER_EMAIL_PROPERTY, DEFAULT_STRING);
```
alikew24, line 89: is the email in the form or is this from the facebook api?

src/main/webapp/facebook-login.js
```diff
@@ -21,35 +21,40 @@ function include(javascriptFile) {
...
 }
-
+ 
 // Called after the user logs in with Facebook
 function checkLoginState() {
-  FacebookAPI.checkLoginStatus(function(response) {
+  FB.getLoginStatus(function(response) {
     statusChangeCallback(response);
   });
 }
```
alikew24, line 60: nit: new line at end of file

src/test/java/com/google/sps/servlets/UserDataServletTest.java
```diff
@@ -0,0 +1,238 @@
...
+    // Build the correct response output
+    Map<String, Object> correctResponse = new HashMap<>();
+    correctResponse.put(USER_FOUND_PROPERTY, false);
+
+    // Check that the user wasn't found in datastore
+    verify(mockRequest, atLeast(1)).getParameter(USER_ID_PROPERTY);
+    Assert.assertEquals(gson.toJson(correctResponse), stringWriter.toString().trim());
+  }
+
+  /** Tests the doGet method, where datastore does contain the test user id */
```
alikew24, line 115: nit: could be helpful to explain what the expected output should be for all these tests

src/main/java/com/google/sps/servlets/UserDataServlet.java
jessdejong: From the facebook API. If we wanted to add it to the form later, we could, but right now the user only provides their email when they log-in.

src/main/webapp/facebook-login.js
jessdejong: Thanks! Good catch.

jessdejong: That's weird, it looks like there is a newline in my code, but not on github. I'll add an extra newline and see if that fixes the problem.

jessdejong: Hmm... there seems to be some bug. Whenever I 'edit' the file on github, there appears to be 2 newlines. However, when I view the file on github, there are no newlines. On my local cloud shell, there is one new line, and git says no new changes are available to commit.

src/test/java/com/google/sps/servlets/UserDataServletTest.java
jessdejong: Sure, I'll do that.

src/main/webapp/facebook-login.js
jessdejong: Not sure what's going on here, I'm leaving this be for now. I think that this can also be fixed when merging (or in another branch).

jessdejong:
@daniel-google @estep-google

src/test/java/com/google/sps/servlets/UserDataServletTest.java
```diff
@@ -0,0 +1,256 @@
...
+    correctResponse.put(USER_FOUND_PROPERTY, false);
+
+    // Check that the user wasn't found in datastore
+    verify(mockRequest, atLeast(1)).getParameter(USER_ID_PROPERTY);
+    Assert.assertEquals(gson.toJson(correctResponse), stringWriter.toString().trim());
+  }
+
+  /**
+   * Tests the doGet method, where datastore does contain the test user id.
+   * Expected output: JSON response with "user-found" property as true,
```
estep-google, line 120: Second paragraph?

src/test/java/com/google/sps/servlets/UserDataServletTest.java
```diff
@@ -0,0 +1,256 @@
...
+  @Test
+  public void testGetMethodWithInvalidId() throws Exception {
+    // Mock the getParameter call
+    when(mockRequest.getParameter(USER_ID_PROPERTY)).thenReturn(TEST_USER_ID);
+
+    // Store output given by mockResponse
+    stringWriter = new StringWriter();
+    writer = new PrintWriter(stringWriter);
+    when(mockResponse.getWriter()).thenReturn(writer);
+    when(mockResponse.getContentType()).thenReturn("application/json");
```
estep-google, line 101: There might be a good reason here, but why do we need to mock the content type? You set it yourself in the doGet(). If you need to check that the content type is correct, you can't do that with this (and in reality you don't need that exact check).

src/test/java/com/google/sps/servlets/UserDataServletTest.java
```diff
@@ -0,0 +1,256 @@
...
+
+  private final Gson gson = new Gson();
+  // Uses a local datastore stored in memory for tests
+  private final LocalServiceTestHelper helper =
+      new LocalServiceTestHelper(new LocalDatastoreServiceTestConfig());
+
+  @Mock private HttpServletRequest mockRequest;
+  @Mock private HttpServletResponse mockResponse;
+  private UserDataServlet servletUnderTest;
+  private StringWriter stringWriter;
```
estep-google, line 73: Are writers something you can set up in the set up? Worth noting is that tests do not share state (except maybe the in-memory database).

src/test/java/com/google/sps/servlets/UserDataServletTest.java
```diff
@@ -0,0 +1,256 @@
...
+
+    // Create a local datastore
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+    addTestUserEntityToDatastore(datastore);
+    // Test the doGet method by calling the wrapper method
+    servletUnderTest.doGetWrapper(datastore, mockRequest, mockResponse);
+    writer.flush();
+
+    // Build the correct response output
+    Map<String, Object> correctResponse = new HashMap<>();
```
estep-google, line 142: ImmutableMap<String, Object> expected = ImmutableMap.of(
    USER_FOUND_PROPERTY, true,
    USER_NAME_PROPERTY, TEST_USER_NAME,
    ...
);

src/test/java/com/google/sps/servlets/UserDataServletTest.java
```diff
@@ -0,0 +1,256 @@
...
+    // Check that the user was found in Datastore
+    verify(mockRequest, atLeast(1)).getParameter(USER_ID_PROPERTY);
+    Assert.assertEquals(gson.toJson(correctResponse), stringWriter.toString().trim());
+  }
+
+  /**
+   * Tests the doPost method, making sure it redirects correctly.
+   * Expected response: Redirects to the profile page with the user id.
+   */
+
```
estep-google, line 159: Remove

src/test/java/com/google/sps/servlets/UserDataServletTest.java
```diff
@@ -0,0 +1,256 @@
...
+  public void testPostMethodRedirect() throws Exception {
+    // Mock the getParameter and getParameterValues calls
+    when(mockRequest.getParameter(USER_ID_PROPERTY)).thenReturn(TEST_USER_ID);
+    when(mockRequest.getParameter(USER_NAME_PROPERTY)).thenReturn(TEST_USER_NAME);
+    when(mockRequest.getParameter(USER_EMAIL_PROPERTY)).thenReturn(TEST_USER_EMAIL);
+    when(mockRequest.getParameter(USER_BIO_PROPERTY)).thenReturn(TEST_USER_BIO);
+    when(mockRequest.getParameterValues(USER_FRIENDS_LIST_PROPERTY)).thenReturn(TEST_USER_FRIENDS_LIST);
+
+    // Create a local datastore
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
```
estep-google, line 170: Random thought: Since you are setting the `servletUnderTest` in the `setUp`, should you maybe do the datastore there too?

src/test/java/com/google/sps/servlets/UserDataServletTest.java
```diff
@@ -0,0 +1,256 @@
...
+    // Test the doPost method by calling the wrapper method
+    servletUnderTest.doPostWrapper(datastore, mockRequest, mockResponse);
+
+    // Query the local datastore for the newly created user entity
```
estep-google, line 197:
```
Entity userEntity = datastore.prepare(new Query(TEST_USER_ID)).asSingleEntity();
assertThat(userEntity).isNotNull();
```

src/test/java/com/google/sps/servlets/UserDataServletTest.java
```diff
@@ -0,0 +1,256 @@
...
+    when(mockRequest.getParameter(USER_BIO_PROPERTY)).thenReturn(TEST_USER_BIO);
+    when(mockRequest.getParameterValues(USER_FRIENDS_LIST_PROPERTY)).thenReturn(TEST_USER_FRIENDS_LIST);
+
+    // Create a local datastore
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+    addTestUserEntityToDatastore(datastore);
+    // Test the doPost method by calling the wrapper method
+    servletUnderTest.doPostWrapper(datastore, mockRequest, mockResponse);
+
+    // Query the local datastore for the newly created user entity
```
estep-google, line 231: Same as above.

src/test/java/com/google/sps/servlets/UserDataServletTest.java
```diff
@@ -0,0 +1,256 @@
...
+    PreparedQuery results = datastore.prepare(new Query(TEST_USER_ID));
+    Assert.assertEquals(1, results.countEntities());
+    Entity userEntity = results.asSingleEntity();
+
+    // Verify that the user Entity name property changed
+    Assert.assertEquals((String) userEntity.getProperty(USER_ID_PROPERTY), TEST_USER_ID);
+    Assert.assertEquals((String) userEntity.getProperty(USER_NAME_PROPERTY), ALTERNATE_TEST_USER_NAME);
```
estep-google, line 238: This is really the only line you need here. You tested all the other lines in the other test case. You can leave the ID check if you want (actually I'd argue that's a good idea).

src/main/java/com/google/sps/servlets/UserDataServlet.java
```diff
@@ -6,6 +6,7 @@
 //
 //     https://www.apache.org/licenses/LICENSE-2.0
 //
+
```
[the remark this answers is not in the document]
jessdejong: whoops! didn't notice that at all.

src/main/java/com/google/sps/servlets/UserDataServlet.java
```diff
@@ -15,18 +16,148 @@
...
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
+import com.google.appengine.api.datastore.DatastoreService;
+import com.google.appengine.api.datastore.DatastoreServiceFactory;
+import com.google.appengine.api.datastore.Entity;
+import com.google.appengine.api.datastore.PreparedQuery;
+import com.google.appengine.api.datastore.Query;
+import com.google.gson.Gson;
+
+/**
```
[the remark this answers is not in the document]
jessdejong: Thanks, didn't know about that. I'll make the change.

src/main/java/com/google/sps/servlets/UserDataServlet.java
```diff
@@ -15,18 +16,148 @@
...
+      userData.put(USER_NAME_PROPERTY, name);
+      userData.put(USER_ID_PROPERTY, id);
+      userData.put(USER_EMAIL_PROPERTY, email);
+      userData.put(USER_BIO_PROPERTY, bio);
+      userData.put(USER_FRIENDS_LIST_PROPERTY, friendsIds);
+    }
+
+    // Send the user's json data as the response
+    response.setContentType("application/json");
+    response.getWriter().println(gson.toJson(userData));
```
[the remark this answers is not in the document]
jessdejong: Thanks! I don't need `println()`, `print()` works.

src/main/java/com/google/sps/servlets/UserDataServlet.java
```diff
@@ -15,18 +16,148 @@
...
+
   @Override
   public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
-    response.setContentType("text/html");
-    response.getWriter().println("Sample response.");
+    // Get the userId, and query Datastore for the corresponding entity
+    String userId = getStringParameter(request, USER_ID_PROPERTY, DEFAULT_STRING);
+    PreparedQuery results = datastore.prepare(new Query(userId));
+
+    Map<String, Object> userData = new HashMap<>();
```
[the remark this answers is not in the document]
jessdejong: Okay, that makes sense. Looks like ImmutableMap is part of the guava framework that Isabelle already added, so I'll go ahead and use that!

src/test/java/com/google/sps/servlets/UserDataServletTest.java
jessdejong: I used an `ImmutableMap`, but used the `put` method for each property. The static `of` method didn't compile, the documentation [here](https://guava.dev/releases/23.0/api/docs/com/google/common/collect/ImmutableMap.html#of--) details that of takes a maximum of 6 key value pairs.

src/main/java/com/google/sps/servlets/UserDataServlet.java
```diff
@@ -15,18 +16,148 @@
...
+    if (results.countEntities() == 0) {
+      // If a user entity was not found
+      userData.put(USER_FOUND_PROPERTY, false);
+    }
+    else {
+      // If a user entity was found (a single entity)
+      Entity userEntity = results.asSingleEntity();
+
+      // Get the user's information
+      String name = (String) userEntity.getProperty(USER_NAME_PROPERTY);
```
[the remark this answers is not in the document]
jessdejong: Yes! It's necessary, the return type of the `getProperty` method is of type `Object`. And I'll put the statements inline, that's more concise. Thanks!

src/main/webapp/profile.html
```diff
@@ -11,58 +11,59 @@
...
             <img src="images/placeholder.jpg" class="rounded-circle" alt="placeholder" width="304" height="236">
           </div>
           <div>
-            <input type="file" accept="image/*" name="photo-profile">
+            <input type="file" accept="image/*" name="profile-photo-1">
           </div>
           <div class="form-group">
             <label for="name">Name:</label>
-            <input type="text" class="form-control" placeholder="Enter name" id="name" width="50">
+            <textarea name="name" type="text" class="form-control" placeholder="Enter name" id="name" width="50" maxlength="300" required></textarea>
```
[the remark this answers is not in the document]
jessdejong: That's better, thanks!. Before I wasn't able to get the input tag to work (for some reason) to send the name parameter to the servlet, but now it works.

src/main/webapp/profile.html
```diff
@@ -11,58 +11,59 @@
...
           </div>
           <div class="form-group">
             <label for="bio">Bio (max 300 characters):</label>
             <textarea name="bio" class="form-control" placeholder="Enter a couple of words about yourself" 
-                id="bio" maxlength="300"></textarea>
+                id="bio" maxlength="300" required></textarea>
```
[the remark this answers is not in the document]
jessdejong: Removed the `required` property from the bio. That makes sense, a bio shouldn't be necessary.

src/main/webapp/facebook-login.js
```diff
@@ -21,36 +21,40 @@ function include(javascriptFile) {
...
-
```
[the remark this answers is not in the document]
jessdejong: Added a newline. I had some problems with this previously (refer to one of Isabelle's comments). Hopefully it's fixed now.

src/test/java/com/google/sps/servlets/UserDataServletTest.java
```diff
@@ -0,0 +1,256 @@
...
+    // Test the doGet method by calling the wrapper method
+    servletUnderTest.doGetWrapper(datastore, mockRequest, mockResponse);
+    writer.flush();
+
+    // Build the correct response output
+    Map<String, Object> correctResponse = new HashMap<>();
+    correctResponse.put(USER_FOUND_PROPERTY, false);
+
+    // Check that the user wasn't found in datastore
+    verify(mockRequest, atLeast(1)).getParameter(USER_ID_PROPERTY);
```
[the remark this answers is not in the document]
jessdejong: This makes sense. I'll remove this verify call, it's unnecessary.

src/test/java/com/google/sps/servlets/UserDataServletTest.java
jessdejong: I think I'll leave that ID check there, just to test one of the properties that are not changing. Also because entities `kind`s are their user ids.

src/test/java/com/google/sps/servlets/UserDataServletTest.java
```diff
@@ -0,0 +1,256 @@
...
+    Assert.assertEquals(gson.toJson(correctResponse), stringWriter.toString().trim());
```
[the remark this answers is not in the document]
jessdejong: Sounds good, Truth looks like it provides better assertions. I'm using the `contains` method for `StringSubject` (returned by `assertThat`) because it looked to be the most suitable method, especially since the mockRequest returns a JSON string.

src/test/java/com/google/sps/servlets/UserDataServletTest.java
jessdejong: Makes sense that it's not necessary. Removed.

src/test/java/com/google/sps/servlets/UserDataServletTest.java
```diff
@@ -0,0 +1,256 @@
...
+  }
+
+  /**
+   * Tests the doGet method, where the datastore doesn't contain the test user id.
+   * Expected output: JSON response with "user-found" property as false.
+   */
+  @Test
+  public void testGetMethodWithInvalidId() throws Exception {
+    // Mock the getParameter call
+    when(mockRequest.getParameter(USER_ID_PROPERTY)).thenReturn(TEST_USER_ID);
```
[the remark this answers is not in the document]
jessdejong: I couldn't find much documentation on how to create the HttpServletRequest and HttpServletResponse objects, and it seems like these objects weren't meant to be created. https://stackoverflow.com/questions/43328707/create-the-httpservletrequest-and-httpservletresponse-message-instead-of-mock.
From the response in that stack overflow post, it seems like trying to create the response and request (apparently requires implementing the interface) would be more effort than it's worth.

src/test/java/com/google/sps/servlets/UserDataServletTest.java
jessdejong: Good point, if tests do not share state, then it wouldn't provide any benefits for creating those private instance variables. I'll declare and instantiate them within the tests.

src/test/java/com/google/sps/servlets/UserDataServletTest.java
jessdejong: That makes sense. I made the change. Unfortunately there doesn't seem to be any way to close the datastore upon tear down, however.

src/main/java/com/google/sps/servlets/UserDataServlet.java
```diff
@@ -58,33 +58,27 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
     String userId = getStringParameter(request, USER_ID_PROPERTY, DEFAULT_STRING);
     PreparedQuery results = datastore.prepare(new Query(userId));
```
estep-google, line 61: Something I think I suggested in the test that I think makes sense here. You would change the `if (count == 0)` check to `if (entity is null)`:

```
Entity entity = datastore.prepare(new Query(userId)).asSingleEntity();
```

src/main/java/com/google/sps/servlets/UserDataServlet.java
```diff
@@ -98,8 +92,8 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
 
     // Check if a user entity with userId already exists
     PreparedQuery results = datastore.prepare(new Query(userId));
```
estep-google, line 96: Chain as mentioned above.

src/test/java/com/google/sps/servlets/UserDataServletTest.java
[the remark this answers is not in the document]
estep-google: Makes sense. Thanks.

src/test/java/com/google/sps/servlets/UserDataServletTest.java
```diff
@@ -107,17 +104,19 @@ public void testGetMethodWithInvalidId() throws Exception {
     writer.flush();
 
     // Build the correct response output
-    Map<String, Object> correctResponse = new HashMap<>();
-    correctResponse.put(USER_FOUND_PROPERTY, false);
+    ImmutableMap<String, Object> expected = ImmutableMap.of(
+        USER_FOUND_PROPERTY, false);
+    //Map<String, Object> correctResponse = new HashMap<>();
```
estep-google: Remove

src/test/java/com/google/sps/servlets/UserDataServletTest.java
```diff
@@ -107,17 +104,19 @@ public void testGetMethodWithInvalidId() throws Exception {
...
+    //correctResponse.put(USER_FOUND_PROPERTY, false);
 
     // Check that the user wasn't found in datastore
-    verify(mockRequest, atLeast(1)).getParameter(USER_ID_PROPERTY);
-    Assert.assertEquals(gson.toJson(correctResponse), stringWriter.toString().trim());
+    assertThat(gson.toJson(expected)).contains(stringWriter.toString());
```
estep-google, line 111: These are strings, so I'm not sure how it works. Is there an "equals" or a "containsExactly" what would work here? Maybe you can convert the output string/json back to a Java object for comparison?

src/test/java/com/google/sps/servlets/UserDataServletTest.java
```diff
@@ -139,24 +137,23 @@ public void testGetMethodWithValidId() throws Exception {
...
 
     // Build the correct response output
-    Map<String, Object> correctResponse = new HashMap<>();
-    correctResponse.put(USER_FOUND_PROPERTY, true);
-    correctResponse.put(USER_NAME_PROPERTY, TEST_USER_NAME);
-    correctResponse.put(USER_ID_PROPERTY, TEST_USER_ID);
-    correctResponse.put(USER_EMAIL_PROPERTY, TEST_USER_EMAIL);
-    correctResponse.put(USER_BIO_PROPERTY, TEST_USER_BIO);
-    correctResponse.put(USER_FRIENDS_LIST_PROPERTY, Arrays.asList(TEST_USER_FRIENDS_LIST));
+    ImmutableMap.Builder<String, Object> expected = ImmutableMap.builder();
```
estep-google, line 136: You can simplify this just a tad:

```
ImmutableMap<String, Object> expected = ImmutableMap.of(
  USER_FOUND_PROPERTY, true,
  USER_BIO_PROPERTY,  TEST_USER_BIO,
  ...
);
```

src/test/java/com/google/sps/servlets/UserDataServletTest.java
```diff
@@ -172,13 +169,13 @@ public void testPostMethodRedirect() throws Exception {
     servletUnderTest.doPostWrapper(datastore, mockRequest, mockResponse);
 
     // Check that the mockResponse redirected property
-    verify(mockRequest, atLeast(1)).getParameter(USER_ID_PROPERTY);
     verify(mockResponse).sendRedirect("/profile.html?id=" + TEST_USER_ID);
```
estep-google, line 166: Ok, sorry, remind me why we expect it to redirect?

src/test/java/com/google/sps/servlets/UserDataServletTest.java
```diff
@@ -195,22 +192,22 @@ public void testPostCreateUserInfo() throws Exception {
...
-    Assert.assertEquals((String) userEntity.getProperty(USER_NAME_PROPERTY), TEST_USER_NAME);
-    Assert.assertEquals((String) userEntity.getProperty(USER_EMAIL_PROPERTY), TEST_USER_EMAIL);
-    Assert.assertEquals((String) userEntity.getProperty(USER_BIO_PROPERTY), TEST_USER_BIO);
-    Assert.assertEquals(((ArrayList<String>) userEntity.getProperty(USER_FRIENDS_LIST_PROPERTY)).toArray(),
-        TEST_USER_FRIENDS_LIST);
+    assertThat((String) userEntity.getProperty(USER_ID_PROPERTY)).contains(TEST_USER_ID);
+    assertThat((String) userEntity.getProperty(USER_NAME_PROPERTY)).contains(TEST_USER_NAME);
+    assertThat((String) userEntity.getProperty(USER_EMAIL_PROPERTY)).contains(TEST_USER_EMAIL);
+    assertThat((String) userEntity.getProperty(USER_BIO_PROPERTY)).contains(TEST_USER_BIO);
+    assertThat(Arrays.toString(((ArrayList<String>) userEntity.getProperty(USER_FRIENDS_LIST_PROPERTY)).toArray()))
```
estep-google, line 195: Why is this going to Strings (I think there's a second usage of this pattern lower)? It seems to me something like this would work:

```
assertThat((ArrayList<String>) userEntity.getProperty(USER_FRIENDS_LIST_PROPERTY).containsExactly(Arrays.toList(TEST_USER_FRIENDS_LIST);
```

src/test/java/com/google/sps/servlets/UserDataServletTest.java
```diff
@@ -229,17 +226,12 @@ public void testPostUpdateUserInfo() throws Exception {
...
+    assertThat(userEntity).isNotNull();
 
     // Verify that the user Entity name property changed
-    Assert.assertEquals((String) userEntity.getProperty(USER_ID_PROPERTY), TEST_USER_ID);
-    Assert.assertEquals((String) userEntity.getProperty(USER_NAME_PROPERTY), ALTERNATE_TEST_USER_NAME);
-    Assert.assertEquals((String) userEntity.getProperty(USER_EMAIL_PROPERTY), TEST_USER_EMAIL);
-    Assert.assertEquals((String) userEntity.getProperty(USER_BIO_PROPERTY), TEST_USER_BIO);
-    Assert.assertEquals(((ArrayList<String>) userEntity.getProperty(USER_FRIENDS_LIST_PROPERTY)).toArray(),
-        TEST_USER_FRIENDS_LIST);
+    assertThat((String) userEntity.getProperty(USER_ID_PROPERTY)).contains(TEST_USER_ID);
```
estep-google, line 223: `.isEqualTo()` - `contains()` (and it's cousins) are for collections. This is a string and you are looking for equality.

src/test/java/com/google/sps/servlets/UserDataServletTest.java
```diff
@@ -229,17 +226,12 @@ public void testPostUpdateUserInfo() throws Exception {
...
+    assertThat((String) userEntity.getProperty(USER_NAME_PROPERTY)).contains(ALTERNATE_TEST_USER_NAME);
```
estep-google, line 224: `isEqualTo()`

pom.xml
```diff
@@ -69,8 +66,18 @@
       <version>1.9.60</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
```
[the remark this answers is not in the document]
jessdejong: I think I actually did that (merge with main), to get the `guava` and `truth` dependencies all I did was resolve merge conflicts from main. I'm not sure why line 69 appears as green, but it's probably because I had to add a couple of lines to resolve a merge conflict. Looking at my  changes it looks like the majority of the guava and truth dependency tags appear to be white (unchanged), so I think that means I took them from a merge with main. I could be wrong though, and I could do a rebase + merge anyway if you'd like.

src/test/java/com/google/sps/servlets/UserDataServletTest.java
jessdejong: The doPost method always redirects to the profile page, because user information is always being updated from either the Login page (in which case they should go to the profile page) or the profile page itself (in which case they should also go to the profile page).

src/test/java/com/google/sps/servlets/UserDataServletTest.java
jessdejong: https://github.com/googleinterns/step48-2020/pull/13#discussion_r452506036. Mentioned this in one of the comments from the previous review, but I resolved it.

src/test/java/com/google/sps/servlets/UserDataServletTest.java
jessdejong: That makes sense.

src/test/java/com/google/sps/servlets/UserDataServletTest.java
jessdejong: [MapSubject documentation](https://truth.dev/api/0.25/com/google/common/truth/MapSubject.html). It seems like I could use the `containsEntry` method for `MapSubjects` in order to compare the mock response to the expected response. This would involve using `Map` (I haven't been able to get `ImmutableMap` to work in this case), and using `gson.fromJson()` to convert the stringwriter's JSON string value. It would look something like this:

`Map<String, Object> actual = (Map) gson.fromJson(stringWriter.toString(), Map.class);
assertThat(expected).containsEntry(USER_FOUND_PROPERTY, actual.get(USER_FOUND_PROPERTY));`

Where there would be an assertion for each property/key. Since isEqualsTo() would be less code, I'm going to stick with that for now, unless you think I should do otherwise.

src/test/java/com/google/sps/servlets/UserDataServletTest.java
jessdejong: Let me know if you think using `of` is still possible!

src/test/java/com/google/sps/servlets/UserDataServletTest.java
estep-google: That deserialization looks OK to me. I don't think you need to assert on all properties? If not, individual is better. If you do/want to, then look at something like `containsExactly()`.

Not being immutable here is fine for the test. I do wonder why you have a `(Map)` case if the `fromJson(...)` is specifying to return a map.

src/test/java/com/google/sps/servlets/UserDataServletTest.java
estep-google: Sorry, I can't find that. Why is `of()` a problem here?

pom.xml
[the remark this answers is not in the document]
estep-google: If there is no substantive change made here, we shouldn't include the file in the PR.

src/test/java/com/google/sps/servlets/UserDataServletTest.java
jessdejong: `of()` takes up to 5 key-value pairs in its argument. Here is a link to the `of()` documentation: [of()](https://guava.dev/releases/23.0/api/docs/com/google/common/collect/ImmutableMap.html#of--)

src/test/java/com/google/sps/servlets/UserDataServletTest.java
jessdejong: `MapSubjects` don't seem to provide the `containsExactly()` method. I changed the code to use `containsEntry()` though.

src/test/java/com/google/sps/servlets/UserDataServletTest.java
```diff
@@ -108,7 +109,8 @@ public void testGetMethodWithInvalidId() throws Exception {
         USER_FOUND_PROPERTY, false);
 
     // Check that the user wasn't found in datastore
-    assertThat(gson.toJson(expected)).isEqualTo(stringWriter.toString());
+    Map<String, Object> actual = gson.fromJson(stringWriter.toString(), Map.class);
+    assertThat(expected).containsEntry(USER_FOUND_PROPERTY, actual.get(USER_FOUND_PROPERTY));
```
estep-google, line 113: General assertion ordering is actual -> expected.

Here, as I've looked a tad more into Truth, we can actually do this (no need for the expected map):

```
assertThat(actual).containsEntry(USER_FOUND_PROPERTY, false);
```

src/test/java/com/google/sps/servlets/UserDataServletTest.java
```diff
@@ -142,7 +144,9 @@ public void testGetMethodWithValidId() throws Exception {
     expected.put(USER_NAME_PROPERTY, TEST_USER_NAME);
 
     // Check that the user was found in Datastore
-    assertThat(gson.toJson(expected.build())).isEqualTo(stringWriter.toString());
+    Map<String, Object> actual = gson.fromJson(stringWriter.toString(), Map.class);
+    assertThat(expected.build()).containsEntry(USER_FOUND_PROPERTY, actual.get(USER_FOUND_PROPERTY));
```
estep-google, line 148: See earlier comment about assertion ordering and not needing the expected map.

jessdejong:
@estep-google Forgot to mention this in the last round of review, but I'm not sure what you mean by removing pom.xml from the pull request, because I did add 3 new dependencies that aren't currently a part of the main branch. The two dependencies that Isabelle added I did take from the main branch though.

src/main/java/com/google/sps/servlets/UserDataServlet.java
```diff
@@ -15,18 +15,138 @@
...
+
+    ImmutableMap.Builder<String, Object> userDataBuilder = ImmutableMap.builder();
+    if (userEntity == null) {
+      // If a user entity was not found
+      userDataBuilder.put(USER_FOUND_PROPERTY, false);
+    }
+    else {
+      // If a user entity was found (a single entity)
+      // Get the user's information
+      userDataBuilder.put(USER_FOUND_PROPERTY, true);
```
estep-google, line 69: You can chain these (`put` returns the builder object):

```
userDataBuilder.put(USER_FOUND_PROPERTY, true)
    .put(USER_BIO_PROPERTY, (String) userEntity.getProperty(USER_BIO_PROPERTY))
    ...
    .put(...);
```

src/main/java/com/google/sps/servlets/UserDataServlet.java
```diff
@@ -15,18 +15,138 @@
...
+      }
+    }
+    datastore.put(userEntity);
+
+    // Redirect to the profile page, and let the front-end know the current logged in user
+    response.sendRedirect("/profile.html?id=" + userId);
+  }
+
+  /** Wrap the doGet method for jUnit testing */
+  public void doGetWrapper(DatastoreService datastoreService, HttpServletRequest request, HttpServletResponse response) throws IOException {
```
estep-google, line 122: New idea - create a package private constructor that takes in the `DatastoreService`. Carefully document that it is for testing purposes only. At that point, you don't need these wrappers and the `datastore` variable can be final (I think - maybe not which is fine too).

You would initialize your fake data store before the servlet in your set up function of the test and be on your way. You'd also be able to remove the member variable for the data store in your test class (you can even probably do it inline in your constructor call).

src/test/java/com/google/sps/servlets/UserDataServletTest.java
```diff
@@ -0,0 +1,228 @@
...
+  private final LocalServiceTestHelper helper =
+      new LocalServiceTestHelper(new LocalDatastoreServiceTestConfig());
+
+  @Mock private HttpServletRequest mockRequest;
+  @Mock private HttpServletResponse mockResponse;
+  private DatastoreService datastore;
+  private UserDataServlet servletUnderTest;
+
+  @Before
+  public void setUp() {
```
estep-google, line 76: For concrete guidance from my comment in the servlet:

```
@Before
public void setUp() {
  MockitoAnnotations.initMocks(this);
  helper.setUp();
  servletUnderTest = new UserDataServlet(DatastoreServiceFactory.getDatastoreService());
}
```

src/test/java/com/google/sps/servlets/UserDataServletTest.java
```diff
@@ -0,0 +1,228 @@
...
+  public void testGetMethodWithValidId() throws Exception {
+    // Mock the getParameter call
+    when(mockRequest.getParameter(USER_ID_PROPERTY)).thenReturn(TEST_USER_ID);
+
+    // Store output given by mockResponse
+    StringWriter stringWriter = new StringWriter();
+    PrintWriter writer = new PrintWriter(stringWriter);
+    when(mockResponse.getWriter()).thenReturn(writer);
+
+    addTestUserEntityToDatastore(datastore);
```
estep-google, line 128: This probably negates my earlier idea of not having a local data store variable

src/main/java/com/google/sps/servlets/UserDataServlet.java
jessdejong: Makes sense, I'll make the change.

jessdejong:
@estep-google Since some of my tests involve checking the test datastore to see if a user is already in the datastore (for example), I'm not sure if I understand how I could eliminate the local datastore from my tests completely... I understand the use of the constructor, but should I pass in the local datastore as the constructor (so that I can call addTestUserEntityToDatastore in my tests, so that the tests work properly)?

Also I was able to replace the wrapper methods with the constructor, but once I did, the servlet stopped working. It's likely because the datastore does not become initialized (because the constructor isn't called). However, even when I tried adding the init() method the constructor still seemed to cause errors. Not sure how to proceed here.

src/test/java/com/google/sps/servlets/UserDataServletTest.java
jessdejong: Confused, how can we eliminate the local data store variable? Seems like it is required in the tests.

src/main/java/com/google/sps/servlets/UserDataServlet.java
jessdejong: See https://github.com/googleinterns/step48-2020/pull/13#issuecomment-656928605

src/test/java/com/google/sps/servlets/UserDataServletTest.java
jessdejong: See https://github.com/googleinterns/step48-2020/pull/13#issuecomment-656928605

jessdejong: (don't understand how it's possible to eliminate the local datastore)

src/test/java/com/google/sps/servlets/UserDataServletTest.java
jessdejong: See https://github.com/googleinterns/step48-2020/pull/13#issuecomment-656928605

estep-google:
You concerns are correct

- In the test you certainly do need access to the data store object (local or member) - keep that as a member variable and create it in set up
- The problem with the package private constructor is probably that you are not calling the default (technically super) constructor (code below) - this is a very quick check
- If a package private constructor does not work, abandon it and do what you had, but those test methods should *not* be public (public methods are considered the API of your class and this is only for unit testing)

```
public class UserDataServlet extends HttpServlet {
  private DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();

  // You may need to have your own default constructor
  // This would be functionally the same as the automatic initialization you have above
  // If the constructor for testing does not work, do not do this
  // If the constructor for testing does work, do not initialize at its declaration site:
  // private *final* DatastoreService datastore;
  public UserDataServlet() {
    super();
    this.datastoreService = DatastoreServiceFactory.getDatastoreService();
  }

  UserDataServlet(DatastoreService datastore) {
    super();  // this is the call that runs the default constructor in HttpServlet
    this.datastore = datastore;
  }

  ...
}
```

jessdejong:
@estep-google Thanks for your help! Unfortunately, adding the default constructor and calling the super method did not work. I ended up making the test wrapper methods package-private.

src/test/java/com/google/sps/servlets/UserDataServletTest.java
```diff
@@ -216,7 +217,8 @@ public void testPostUpdateUserInfo() throws Exception {
 
   /** Helper method to add a test user to the local datastore */
   private void addTestUserEntityToDatastore(DatastoreService datastore) {
-    Entity userEntity = new Entity(TEST_USER_ID);
+    //Entity userEntity = new Entity(TEST_USER_ID);
+    Entity userEntity = new Entity("User");
```
estep-google, line 221: I would assume you'd want `new Entity(USER_ENTITY);`?

jessdejong: Good catch, I'll fix that!

PR #19, opened 2020-07-10
Add Methods and Servlet for Finding Potential Matches

mivdittamo:

- Add methods for finding the potential matches for a user

- Add servlet to find and return the next potential match for a user's feed page

- Tests added to check the functionality of both the potential match methods and the servlet

### Is this fixing an issue?

none

- Currently, the PotentialMatchesServlet is directly running the potential match finding methods and finding all of the potential matches for a user every time a get request is made. Eventually, this will be changed so that the finding methods are run separately with the resulting potential matches in put datastore, and then this servlet will obtain that list from datastore.

- [ X ] Test written/updating
- [ X ] Tests passing
- [ X ] Coding style (indentation, etc)


src/main/java/com/google/sps/data/potentialMatchAlgorithm.java
```diff
@@ -0,0 +1,69 @@
...
+import com.google.common.collect.ImmutableMap;
+import com.google.sps.data.friend_map.UserFriendsMap;
+import com.google.sps.data.friend_map.UserNode;
+
+/**
+* Class that contains methods that find the potential matches for users.
+*
+* <p>Potential matches are users with who a particular user shares at least one friend with.
+*/
+public class potentialMatchAlgorithm {
```
jessdejong, line 28: nit: should the class name be capitalized?

src/main/java/com/google/sps/data/potentialMatchAlgorithm.java
```diff
@@ -0,0 +1,69 @@
...
+  public static ImmutableSet<String> findPotentialMatchesForUser(String userID, UserFriendsMap friendsMap) {
+    ImmutableSet<String> userFriendIDs = friendsMap.getUserFriendIDs(userID);
+
+    ImmutableSet<String> potentialMatchesIDs;
+    if (userFriendIDs != null) {
+      potentialMatchesIDs = userFriendIDs
+      .stream()
+      .flatMap(friendID -> friendsMap.getUserFriendIDs(friendID).stream())
+      .filter(potentialMatchID -> !potentialMatchID.equals(userID) && !userFriendIDs.contains(potentialMatchID))
+      .collect(ImmutableSet.toImmutableSet());
```
jessdejong, line 62: nit: it looks like you used two spaces for the indentation on your previous use of streams above, maybe keep it consistent.

src/main/java/com/google/sps/servlets/PotentialMatchesServlet.java
```diff
@@ -0,0 +1,109 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.sps.servlets;
+
+import java.util.List;
```
jessdejong, line 17: nit: Alphabetized the import statements (super nit picky haha)

src/main/java/com/google/sps/servlets/PotentialMatchesServlet.java
```diff
@@ -0,0 +1,109 @@
...
+import com.google.appengine.api.datastore.DatastoreService;
+import com.google.appengine.api.datastore.DatastoreServiceFactory;
+import com.google.appengine.api.datastore.Query;
+import com.google.appengine.api.datastore.PreparedQuery;
+import com.google.appengine.api.datastore.FetchOptions;
+import com.google.appengine.api.datastore.Entity;
+import com.google.common.collect.ImmutableSet;
+
+/**
+* Handles requests for getting the next potential match for a user's feed page.
```
jessdejong, line 36: nit: add a space

src/main/java/com/google/sps/servlets/PotentialMatchesServlet.java
```diff
@@ -0,0 +1,109 @@
...
+ */
+@WebServlet("/potential-matches")
+public class PotentialMatchesServlet extends HttpServlet {
+  private static final String USER_ENTITY = "User";
+  private static final String USER_ID_PROPERTY = "id";
+  private static final String USER_FRIENDS_LIST_PROPERTY = "friends-list";
+
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    String currUserID = request.getParameter("userid");
```
jessdejong, line 46: Should the parameter String here have the same value as the USER_ID_PROPERTY? That's how I've been doing it in my code at least.

src/main/java/com/google/sps/servlets/PotentialMatchesServlet.java
```diff
@@ -0,0 +1,109 @@
...
+  * @return The next potential match for a user
+  */
+  private String loadUserPotentialMatch(String userID) {
+    ImmutableSet<UserNode> userNodes = createUserNodes();
+    UserFriendsMap friendsMap = new UserFriendsMap(userNodes);
+    
+    ImmutableSet<String> potentialMatches = potentialMatchAlgorithm.findPotentialMatchesForUser(userID, friendsMap);
+  
+    String nextPotentialMatchID = potentialMatches.size() > 0 ?
+      potentialMatches.iterator().next() : "NO_POTENTIAL_MATCHES";
```
jessdejong, line 72: Make a final variable for this string?

src/main/java/com/google/sps/servlets/PotentialMatchesServlet.java
```diff
@@ -0,0 +1,109 @@
...
+
+  /**
+  * Helper method that creates the set of UserNodes that will be fed into the potential
+  * matching method
+  *
+  * @return the set of user nodes for all current app users
+  */
+  private ImmutableSet<UserNode> createUserNodes() {
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+    PreparedQuery results = datastore.prepare(new Query(USER_ENTITY));
```
jessdejong, line 85: Possibly: `DatastoreServiceFactory.getDatastoreService().prepare(new Query(USER_ENTITY));`
However, the `kind` for user entities, at least for the user entities that I'm storing, is their user id.

jessdejong: So each `kind` corresponds to a single user entity.

src/main/java/com/google/sps/servlets/PotentialMatchesServlet.java
```diff
@@ -0,0 +1,109 @@
...
+    List<Entity> entityResults = results.asList(FetchOptions.Builder.withLimit(Integer.MAX_VALUE));
+
+    ImmutableSet.Builder<UserNode> builder = ImmutableSet.builder();
+    for (Entity userEntity: entityResults) {
```
jessdejong, line 89: nit: space after `userEntity:`

src/main/java/com/google/sps/servlets/PotentialMatchesServlet.java
```diff
@@ -0,0 +1,109 @@
...
+    return allUserNodes;
+  }
+
+/** Wrap the doGet method for jUnit testing */
+  public void doGetWrapper(DatastoreService datastoreService, HttpServletRequest request, HttpServletResponse response) throws IOException {
+    DatastoreService datastore = datastoreService;
+    doGet(request, response);
+  }
+}
+
```
jessdejong, line 109: Nik ended up suggesting to use a package-private constructor that I'm currently still working on. I'll keep you updated.

src/main/java/com/google/sps/data/potentialMatchAlgorithm.java
mivdittamo: Yes, it should be! I'll update that. I also ended up doing this for the test file, so I'll update that too.

src/main/java/com/google/sps/data/potentialMatchAlgorithm.java
mivdittamo: got it!

src/main/java/com/google/sps/servlets/PotentialMatchesServlet.java
mivdittamo: I reordered some of the imports but didn't strictly alphabetize them, but I reorganized them so that related imports are grouped together.

src/main/java/com/google/sps/servlets/PotentialMatchesServlet.java
mivdittamo: got it!

src/main/java/com/google/sps/servlets/PotentialMatchesServlet.java
mivdittamo: The "userid" here refers to the url parameter that's in the request to this servlet, so it's separate from the USER_ID_PROPERTY.

src/main/java/com/google/sps/servlets/PotentialMatchesServlet.java
mivdittamo: Good idea!

src/main/java/com/google/sps/servlets/PotentialMatchesServlet.java
mivdittamo: Discussed and resolved offline

src/main/java/com/google/sps/servlets/PotentialMatchesServlet.java
mivdittamo: Ok! I'll keep this open if there's any updates!

src/test/java/com/google/sps/data/PotentialMatchAlgorithmTest.java
```diff
@@ -0,0 +1,186 @@
...
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+import com.google.common.collect.ImmutableSet;
+import com.google.common.collect.ImmutableMap;
+import com.google.sps.data.friend_map.UserFriendsMap;
+import com.google.sps.data.friend_map.UserNode;
+import com.google.sps.data.PotentialMatchAlgorithm;
+
+/* NOTE: PM = Potential Match .*/
+
```
jessdejong, line 31: nit: remove newline?

mivdittamo: Got it!

src/test/java/com/google/sps/servlets/PotentialMatchesServletTest.java
```diff
@@ -0,0 +1,233 @@
...
+    addTestUserEntityToDatastore(datastore, TEST_USER_1_ID, TEST_USER_1_NAME,
+      TEST_USER_1_EMAIL, TEST_USER_1_BIO, new String[]{});
+    
+    servletUnderTest.doGetWrapper(datastore, mockRequest, mockResponse);
+    writer.flush();
+
+    String expectedOutput = NO_POTENTIAL_MATCH_RESULT;
+    String actualOutput = stringWriter.toString();
+
+    assertThat(actualOutput).isEqualTo(expectedOutput);
```
jessdejong, line 120: Probably not necessary to store these values in a String.

mivdittamo: I put these in separate String variables just for readability.

src/main/java/com/google/sps/servlets/PotentialMatchesServlet.java
```diff
@@ -0,0 +1,110 @@
...
+    }
+
+    ImmutableSet<UserNode> allUserNodes = builder.build();
+    
+    return allUserNodes;
+  }
+
+/** Wrap the doGet method for jUnit testing */
+  public void doGetWrapper(DatastoreService datastoreService, HttpServletRequest request, HttpServletResponse response) throws IOException {
+    DatastoreService datastore = datastoreService;
```
estep-google, line 106: How is this working? You're manually creating your `DatastoreService` inside of `createUserNode()`, so it seems like this wouldn't be used.

src/test/java/com/google/sps/data/PotentialMatchAlgorithmTest.java
```diff
@@ -0,0 +1,185 @@
...
+  private static final String USER_D_ID = "45678";
+  private static final String USER_E_ID = "56789";
+
+  /**
+  * Tests if correct potential matches are found for a single user with zero friends
+  *
+  * <p>Should result in a map with a single key that maps to an empty set.
+  */
+  @Test
+  public void oneUserNoFriendsPMTest() {
```
estep-google, line 46: Naming wise, you don't need to end methods with `Test`. This is indicated by the `@Test` annotation.

Regarding `PM`, I think you can just say Match(es). e.g., `oneUserNoMatches()` or `threeUsersOneMatch()`

src/test/java/com/google/sps/data/PotentialMatchAlgorithmTest.java
```diff
@@ -0,0 +1,185 @@
...
+    UserNode userA = new UserNode(USER_A_ID, ImmutableSet.of());
+
+    Set<UserNode> oneUserNodeSet = ImmutableSet.of(userA);
+    UserFriendsMap resultingFriendsMap = new UserFriendsMap(oneUserNodeSet);
+    ImmutableMap<String, ImmutableSet<String>> resultingPotentialMatches =
```
estep-google, line 51: `resultingPotentialMatches` -> `potentialMatches`

src/test/java/com/google/sps/data/PotentialMatchAlgorithmTest.java
```diff
@@ -0,0 +1,185 @@
...
+  @Test
+  public void oneUserNoFriendsPMTest() {
+    UserNode userA = new UserNode(USER_A_ID, ImmutableSet.of());
+
+    Set<UserNode> oneUserNodeSet = ImmutableSet.of(userA);
+    UserFriendsMap resultingFriendsMap = new UserFriendsMap(oneUserNodeSet);
+    ImmutableMap<String, ImmutableSet<String>> resultingPotentialMatches = 
+      PotentialMatchAlgorithm.findAllPotentialMatches(resultingFriendsMap);
+    
+    ImmutableMap<String, ImmutableSet<String>> expectedMap = ImmutableMap.of(
```
estep-google, line 54: `expectedMap` -> `expectedMatches`

src/test/java/com/google/sps/data/PotentialMatchAlgorithmTest.java
```diff
@@ -0,0 +1,185 @@
...
+import org.junit.runners.JUnit4;
+import com.google.common.collect.ImmutableSet;
+import com.google.common.collect.ImmutableMap;
+import com.google.sps.data.friend_map.UserFriendsMap;
+import com.google.sps.data.friend_map.UserNode;
+import com.google.sps.data.PotentialMatchAlgorithm;
+
+/* NOTE: PM = Potential Match .*/
+@RunWith(JUnit4.class)
+public class PotentialMatchAlgorithmTest {
```
estep-google, line 32: Nice and comprehensive test cases!

src/test/java/com/google/sps/servlets/PotentialMatchesServletTest.java
```diff
@@ -0,0 +1,233 @@
...
+    helper.tearDown();
+  }
+
+  /**
+  * Tests scenario where there is only one user in datastore
+  *
+  * <p>Should result in no next potential match found
+  */
+  @Test
+  public void oneUserNoPotentialMatches() throws Exception {
```
estep-google, line 104: Same naming advice. You don't really need Potential here -> `oneUserNoMatches`

src/test/java/com/google/sps/servlets/PotentialMatchesServletTest.java
estep-google: I agree with having the `actualOutput` (`stringWriter.toSring()` is unclear), but the expected can be inlined. `isEqualTo` implies it's the expected.

src/test/java/com/google/sps/servlets/PotentialMatchesServletTest.java
```diff
@@ -0,0 +1,233 @@
...
+    when(mockRequest.getParameter("userid")).thenReturn(TEST_USER_1_ID);
+
+    StringWriter stringWriter = new StringWriter();
+    PrintWriter writer = new PrintWriter(stringWriter);
+    when(mockResponse.getWriter()).thenReturn(writer);
```
estep-google, line 109: This exact block exists in each test. It seems like `stringWriter` should be a member variable (maybe name it `responseWriter` - something like that). This block should be in `setUp()` and you can change the block to this:

```
responseWriter = new StringWriter();
PrintWriter writer = new PrintWriter(responseWriter, /* autoFlush= */ true);
when(mockResponse.getWriter()).thenReturn(writer);
```

By using `autoFlush`, you don't really need to keep track of the `writer` variable in tests.

src/test/java/com/google/sps/servlets/PotentialMatchesServletTest.java
```diff
@@ -0,0 +1,233 @@
...
+  }
+
+  /**
+  * Tests scenario where there is only one user in datastore
+  *
+  * <p>Should result in no next potential match found
+  */
+  @Test
+  public void oneUserNoPotentialMatches() throws Exception {
+    when(mockRequest.getParameter("userid")).thenReturn(TEST_USER_1_ID);
```
estep-google, line 105: `"userid"` should be a static final at the top of the file. It help prevent a typo in a new test case that costs hours of "banging your head against the wall" debugging :)

src/test/java/com/google/sps/servlets/PotentialMatchesServletTest.java
```diff
@@ -0,0 +1,233 @@
...
+
+    StringWriter stringWriter = new StringWriter();
+    PrintWriter writer = new PrintWriter(stringWriter);
+    when(mockResponse.getWriter()).thenReturn(writer);
+
+    addTestUserEntityToDatastore(datastore, TEST_USER_1_ID, TEST_USER_1_NAME,
+      TEST_USER_1_EMAIL, TEST_USER_1_BIO, new String[]{});
+    
+    servletUnderTest.doGetWrapper(datastore, mockRequest, mockResponse);
```
estep-google, line 114: I think a general test structure would be fairly clean like this:

```
public void twoUserNameThings() {
  addTestUserEntityToDatastore(...);
  addTestUserEntityToDatastore(...);  // as needed, of course

  String nextMatch = execute(TEST_USER_1_ID);

  assertThat(nextMatch).isEqualTo(NO_POTENTIAL_MATCH_RESULT);
}

private String execute(String userIdToFetch) {
  when(mockRequest.getParameter(USER_ID_URL_PARAM)).thenReturn(userIdToFetch);

  servletUnderTest.doGetWrapper(datastore, mockRequest, mockResponse);
  return responseWriter.toString();
}
```

src/main/java/com/google/sps/data/PotentialMatchAlgorithm.java
```diff
@@ -0,0 +1,69 @@
...
+  *
+  * @param friendsMap object that contains the map of all direct friendships between all users
+  * @return The map of each user ID to a set of the user IDs of all their potential matches
+  */
+  public static ImmutableMap<String, ImmutableSet<String>> findAllPotentialMatches(UserFriendsMap friendsMap) {
+    Set<String> allUserIDs = friendsMap.getUserIDs();
+
+    ImmutableMap<String, ImmutableSet<String>> allPotentialMatches = allUserIDs
+      .stream()
+      .collect(ImmutableMap.toImmutableMap(id -> id, id -> findPotentialMatchesForUser(id, friendsMap)));
```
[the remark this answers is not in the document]
mivdittamo: Got it!

src/main/java/com/google/sps/data/PotentialMatchAlgorithm.java
```diff
@@ -0,0 +1,69 @@
...
+  *
+  * @param userID The user ID of the user who's potential matches are being found
+  * @param friendMap The map of the direct friendships between all users
+  * @return The set of user IDs of the potential matches that are found
+  */
+  public static ImmutableSet<String> findPotentialMatchesForUser(String userID, UserFriendsMap friendsMap) {
+    ImmutableSet<String> userFriendIDs = friendsMap.getUserFriendIDs(userID);
+
+    ImmutableSet<String> potentialMatchesIDs;
+    if (userFriendIDs != null) {
```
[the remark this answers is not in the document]
mivdittamo: The userFriendIDs would only be null if any of the IDs passed into getUserFriendIDs() were not found, so I added the null check. But I can change this so that getUserFriendIDs() returns an empty set instead.

src/main/java/com/google/sps/data/PotentialMatchAlgorithm.java
```diff
@@ -0,0 +1,69 @@
...
+      potentialMatchesIDs = userFriendIDs
+        .stream()
+        .flatMap(friendID -> friendsMap.getUserFriendIDs(friendID).stream())
+        .filter(potentialMatchID -> !potentialMatchID.equals(userID) && !userFriendIDs.contains(potentialMatchID))
```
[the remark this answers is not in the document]
mivdittamo: We can do the filtering in two steps. Definitely looks a lot cleaner.

src/main/java/com/google/sps/data/PotentialMatchAlgorithm.java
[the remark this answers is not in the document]
mivdittamo: Update to this, changed the getUserFriendIDs() to use the getOrDefault() method to return either the set of friends or an empty set if the id is not found.

src/main/java/com/google/sps/servlets/PotentialMatchesServlet.java
```diff
@@ -0,0 +1,110 @@
...
+  private static final String NO_POTENTIAL_MATCH_RESULT = "NO_POTENTIAL_MATCHES";
+
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    String currUserID = request.getParameter("userid");
+    
+    String nextPotentialMatchID = loadUserPotentialMatch(currUserID);
+
+    response.setContentType("text/html");
+    response.getWriter().print(nextPotentialMatchID);
```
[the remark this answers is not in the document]
mivdittamo: The text returned here would be the next potential match id, but it does seem like a better idea to have this as a JSON object in case we need to later add in more data.

src/main/java/com/google/sps/servlets/PotentialMatchesServlet.java
mivdittamo: Got it!

src/main/java/com/google/sps/servlets/PotentialMatchesServlet.java
```diff
@@ -0,0 +1,110 @@
...
+  * @param userID The userID of the user who's potential match is being found
+  * @return The next potential match for a user
+  */
+  private String loadUserPotentialMatch(String userID) {
+    ImmutableSet<UserNode> userNodes = createUserNodes();
+    UserFriendsMap friendsMap = new UserFriendsMap(userNodes);
+    
+    ImmutableSet<String> potentialMatches = PotentialMatchAlgorithm.findPotentialMatchesForUser(userID, friendsMap);
+  
+    String nextPotentialMatchID = potentialMatches.size() > 0 ?
```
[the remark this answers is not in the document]
mivdittamo: Is it ok to write this as `potentialMatches.iterator().hasNext()`? I can't seem to figure out how to use the method reference way here.

src/main/java/com/google/sps/servlets/PotentialMatchesServlet.java
```diff
@@ -0,0 +1,110 @@
...
+      potentialMatches.iterator().next() : NO_POTENTIAL_MATCH_RESULT;
```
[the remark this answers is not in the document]
mivdittamo: Yes, the current code is only returning the first ID in potentialMatches. I'm currently working on a way to track which IDs have already appeared on a user's feed page so that the next IDs in the potentialMatches list can be returned, and I will push that update when it's completed.

src/main/java/com/google/sps/servlets/PotentialMatchesServlet.java
mivdittamo: Oh you're right! I was following along with Jess's servlet testing to help write the servlet tests for this and used a doGetWrapper method, but realizing now with the way I wrote the methods in my servlet that I don't need to keep the wrapper.

src/test/java/com/google/sps/data/PotentialMatchAlgorithmTest.java
mivdittamo: We'll be adding the functionality to create matches in the near future (both users choose to friend each other on their respective feed pages), and these differ from the potential matches that are found here. To avoid later confusion with the eventual matches, I had put a reference to potential matches here. Should this still be renamed?

src/test/java/com/google/sps/data/PotentialMatchAlgorithmTest.java
mivdittamo: Got it!

src/test/java/com/google/sps/data/PotentialMatchAlgorithmTest.java
mivdittamo: In reference to the comments above, to keep the difference between potential matches and matches, would `expectedPotentialMatches` also be reasonable to rename this variable to?

src/test/java/com/google/sps/data/PotentialMatchAlgorithmTest.java
mivdittamo: Thanks!

src/test/java/com/google/sps/servlets/PotentialMatchesServletTest.java
mivdittamo: Got it!

src/test/java/com/google/sps/servlets/PotentialMatchesServletTest.java
mivdittamo: Makes sense! I'll change the tests to use this.

src/test/java/com/google/sps/servlets/PotentialMatchesServletTest.java
mivdittamo: Got it!

src/test/java/com/google/sps/servlets/PotentialMatchesServletTest.java
mivdittamo: Got it! I had made some modifications to the servlet response so that it returns a JSON object, and I also removed the doGetWrapper, so I used this format you suggested but with some changes to account for these modifications.

src/main/java/com/google/sps/servlets/PotentialMatchesServlet.java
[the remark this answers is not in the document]
estep-google: Sorry, the method reference was just my pointing to the method. What you have is correct.

src/test/java/com/google/sps/data/PotentialMatchAlgorithmTest.java
estep-google: Would you be testing both of those things in this unit test, though? Generating potential matches seems different than putting a match together.

src/test/java/com/google/sps/data/PotentialMatchAlgorithmTest.java
estep-google: It would be better if the earlier comment is resolved in favor of keeping potential match naming.

src/main/java/com/google/sps/servlets/PotentialMatchesServlet.java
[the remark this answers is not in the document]
mivdittamo: Ok, thanks!

src/test/java/com/google/sps/data/PotentialMatchAlgorithmTest.java
mivdittamo: I'll be testing the match creation separately. Since this is the case, I'll update the names for the tests here.

src/test/java/com/google/sps/data/PotentialMatchAlgorithmTest.java
mivdittamo: Updating this name to expectedMatches (reference to earlier comment).

src/test/java/com/google/sps/servlets/PotentialMatchesServletTest.java
mivdittamo: Got it!

src/test/java/com/google/sps/servlets/PotentialMatchesServletTest.java
```diff
@@ -0,0 +1,222 @@
...
+  private String execute(String userIDToFetch) throws IOException{
+    when(mockRequest.getParameter(USER_ID_REQUEST_PARAM)).thenReturn(userIDToFetch);
+
+    servletUnderTest.doGet(mockRequest, mockResponse);
+    
+    String result = responseWriter.toString();
+    JSONObject jsonResponse = new JSONObject(result);
+    String output = jsonResponse.getString(MATCHINFO_NEXT_MATCH_ID_FIELD);
+
+    return output;
```
estep-google, line 209: Inline the return "block":

```
String result = responseWriter.toString();
JSONObject jsonResponse = new JSONObject(result);
return jsonResponse.getString(MATCHINFO_NEXT_MATCH_ID_FIELD);
```

src/main/java/com/google/sps/data/friend_map/UserFriendsMap.java
```diff
@@ -54,7 +54,7 @@ public UserFriendsMap (Set<UserNode> userNodes) {
   }
 
   public ImmutableSet<String> getUserFriendIDs(String currentUser) {
-    return friendMap.get(currentUser);
+    return friendMap.getOrDefault(currentUser, ImmutableSet.of());
```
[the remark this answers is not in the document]
mivdittamo: Got it!

src/main/java/com/google/sps/servlets/PotentialMatchesServlet.java
```diff
@@ -0,0 +1,111 @@
...
+/**
+*  Handles requests for getting the next potential match for a user's feed page.
+*/
+@WebServlet("/potential-matches")
+public class PotentialMatchesServlet extends HttpServlet {
+  private static final String USER_ENTITY = "User";
+  private static final String USER_ID_PROPERTY = "id";
+  private static final String USER_FRIENDS_LIST_PROPERTY = "friends-list";
+  private static final String NO_POTENTIAL_MATCH_RESULT = "NO_POTENTIAL_MATCHES";
+  private static final String USER_ID_REQUEST_PARAM = "userid";
```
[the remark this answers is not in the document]
mivdittamo: Got it!

src/main/java/com/google/sps/servlets/PotentialMatchesServlet.java
```diff
@@ -0,0 +1,111 @@
...
+  private String loadUserPotentialMatch(String userID) {
+    ImmutableSet<UserNode> userNodes = createUserNodes();
+    UserFriendsMap friendsMap = new UserFriendsMap(userNodes);
+    
+    ImmutableSet<String> potentialMatches = PotentialMatchAlgorithm.findPotentialMatchesForUser(userID, friendsMap);
+  
+    String nextPotentialMatchID = potentialMatches.iterator().hasNext() ?
+      potentialMatches.iterator().next() : NO_POTENTIAL_MATCH_RESULT;
+    
+    return nextPotentialMatchID;
```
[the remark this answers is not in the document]
mivdittamo: Got it!

src/main/java/com/google/sps/servlets/PotentialMatchesServlet.java
```diff
@@ -0,0 +1,111 @@
...
+      ArrayList<String> friendsIds =
+        (ArrayList<String>) userEntity.getProperty(USER_FRIENDS_LIST_PROPERTY);
+      UserNode userNode = new UserNode(userID,
+        friendsIds != null ? ImmutableSet.copyOf(friendsIds) : ImmutableSet.of());
+      builder.add(userNode);
+    }
+
+    ImmutableSet<UserNode> allUserNodes = builder.build();
+    
+    return allUserNodes;
```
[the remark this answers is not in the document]
mivdittamo: Got it!

src/test/java/com/google/sps/servlets/PotentialMatchesServletTest.java
mivdittamo: Got it!

PR #25, opened 2020-07-15
adjust for the potential matches response being sent as a json

alikew24:

I am changing the response I am fetching to adjust for it now being sent as a json

### Is this fixing an issue?

none

none

- [  ] Test written/updating
- [  ] Tests passing

### DIFF
--- a/src/main/webapp/script.js
+++ b/src/main/webapp/script.js
@@ -122,15 +122,15 @@ function createIndicator(dataSlideNumber, className) {
 function getNextPotentialMatch() {
   deletePotentialMatchInfo();
   let currentUser = getCurrentUserId();
-  fetch('/potential-matches?userid=' + currentUser).then(response => response.text()).then((pmID) => { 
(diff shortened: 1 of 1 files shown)


src/main/webapp/script.js
```diff
@@ -122,15 +122,15 @@ function createIndicator(dataSlideNumber, className) {
 function getNextPotentialMatch() {
   deletePotentialMatchInfo();
   let currentUser = getCurrentUserId();
-  fetch('/potential-matches?userid=' + currentUser).then(response => response.text()).then((pmID) => { 
+  fetch('/potential-matches?userid=' + currentUser).then(response => response.json()).then((pmID) => { 
       let noMatch = "NO_POTENTIAL_MATCHES";
```
daniel-google, line 126: nit: pull this into a `const` defined at/near the top of the file and add a comment that it must be kept in-sync with XServlet.java

daniel-google: (if the servlet isn't merged yet, feel free to address later when it is)

PR #28, opened 2020-07-20
Load Match Information into Datastore

### LINKED ISSUE #21: Update PotentialMatchesServlet to obtain potential matches from datastore
### Issue summary

Currently the PotentialMatchesServlet is running the potential matching methods directly, and it is finding all of the potential matches for a particular user, and then it returns a single ID from the resulting list. Repeatedly running the potential match finding methods is not very efficient, so eventually these processes will be run separately, and their results will be put in datastore.

### Expected behavior

Run the potential match finding methods in a class separate from the PotentialMatchesServlet, and store the list of results into datastore. Then the PotentialMatchesServlet can retrieve this list to obtain the next potential match.

### Actual behavior

The PotentialMatchesServlet is directly running the potential match finding methods, and one (if any) id is returned as the next potential match.

### Steps to reproduce (including any preconditions)

Making a get request to the PotentialMatchesServlet

### Your environment

Include version numbers

Friends of friends v. 1.

mivdittamo:

- Update the PotentialMatchesServlet to store a user's match information in datastore and only run the potential match algorithm when this information is initially stored

- Retrieve next potential match from stored match information in datastore

- Update PotentialMatchesServlet tests to account for storing information in datastore and also to check the results of multiple calls to this servlet

### Is this fixing an issue?

resolves #21 

- Match information includes the list of a user's potential matches, the list of users they 'friended', and the list of users they have 'passed'

- The pom.xml edits were deletions made due to duplicate dependencies that were created when the main branch was merged with this branch

- [ x ] Test written/updating
- [ x ] Tests passing


src/test/java/com/google/sps/servlets/PotentialMatchesServletTest.java
```diff
@@ -186,9 +207,21 @@ public void fourUsersOneMutualConnection() throws Exception {
     addTestUserEntityToDatastore(datastore, TEST_USER_4_ID, TEST_USER_4_NAME,
       TEST_USER_4_EMAIL, TEST_USER_4_BIO, testUser4FriendsList);
 
-    String actualOutput = execute(TEST_USER_4_ID);
+    String actualOutput_1 = execute(TEST_USER_4_ID);
+    String expected_2 = actualOutput_1.equals(TEST_USER_2_ID) ? TEST_USER_3_ID : TEST_USER_2_ID;
```
alikew24, line 211: nit: why is this expected_2? I don't see an expected_1

src/main/java/com/google/sps/servlets/PotentialMatchesServlet.java
```diff
@@ -60,28 +68,74 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
+    }
+  }
+  
+  /**
+  * Given a specific user, get their next potential match for their feed page
+  *
+  * @param userID The user whose potential match is being retrieved
+  */
+  private String getNextPotentialMatchID(String userID) {
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
```
alikew24, line 95: nit: could this be declared above all the functions with the other variables?

src/test/java/com/google/sps/servlets/PotentialMatchesServletTest.java
mivdittamo: I meant to have this as the expected output for the second call to the servlet's doGet method. I can change this variable to be named expectedSecondOutput to clarify what it's for!

src/main/java/com/google/sps/servlets/PotentialMatchesServlet.java
mivdittamo: Yup, I can make this a final class variable.

pom.xml
```diff
@@ -77,36 +77,6 @@
       <version>1.0.1</version>
       <scope>test</scope>
     </dependency>
-    <dependency>
-    <groupId>org.mockito</groupId>
-      <artifactId>mockito-all</artifactId>
-      <version>1.9.5</version>
-      <scope>test</scope>
```
daniel-google, line 84: Hmm looks like we're still using these dependencies; was this removal intentional?

src/main/java/com/google/sps/servlets/PotentialMatchesServlet.java
```diff
@@ -60,28 +70,69 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
+  * already in datastore.
   *
-  * <p>Currently is just loading all the possible matches and obtaining one of them.
+  * <p>Creates new match information entity for user if it had not already been stored.
   *
-  * <p>TODO(#21): Obtain potential match results from datastore instead of running the 
-  * potential match finding methods repeatedly.
+  * @param userID The ID of the user who's match information is being checked/loaded into datastore
+  */
+  private void loadUserMatchInformation(String userID) {
```
daniel-google, line 80: Since the effect of this method is initializing some data if it does not yet exist, and it doesn't return anything itself, I would either:

1) Change this method to return something meaningful, e.g. the `List<String>` of potential matches, or just the `Entity` itself
2) OR, rename this method to something more descriptive of its intent, e.g. `ensureUserHasMatchInformation`

Returning the Entity or matches directly (option 1) seems more useful, since you only need to load it once in the case where it already exists, and don't need to load it at all in the case where it's newly created.

src/main/java/com/google/sps/servlets/PotentialMatchesServlet.java
```diff
@@ -60,28 +70,69 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
+      addMatchInfoToDatastore(userID);
+    }
+  }
+  
+  /**
+  * Given a specific user, get their next potential match for their feed page
+  *
+  * @param userID The user whose potential match is being retrieved
+  */
+  private String getNextPotentialMatchID(String userID) {
```
daniel-google, line 94: Since this method has a side-effect of updating the list of matches, calling it `get*` may be misleading (as getters are generally side-effect-free).

I would rename it to something like `advanceToNextPotentialMatch` to make the side-effect more clear, or else split into 2 methods e.g. `getFirstPotentialMatch` and `consumeFirstPotentialMatch`.

src/main/java/com/google/sps/servlets/PotentialMatchesServlet.java
```diff
@@ -60,28 +70,69 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
+    Entity matchInfoEntity = datastore.prepare(new Query(MATCH_INFO_ENTITY).setFilter(
+      new FilterPredicate(USER_ID_PROPERTY, FilterOperator.EQUAL, userID))).asSingleEntity();
+
+    List<String> potentialMatches = (ArrayList<String>) matchInfoEntity.getProperty(POTENTIAL_MATCHES_PROPERTY);
```
daniel-google, line 98: nit: there's no reason to assume that `getProperty()` returns a specific `List` implementation here; just cast to `List<String>`

src/main/java/com/google/sps/servlets/PotentialMatchesServlet.java
```diff
@@ -60,28 +70,69 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
+
+    String nextPotentialMatchID;
+    
+    if (potentialMatches != null && potentialMatches.iterator().hasNext()) {
```
daniel-google, line 102: nit: instead of using an Iterator here, it may be more clear to just deal with the 1st element specifically, especially since the way the code works currently is a bit unclear -- i.e. does calling `remove()` on the iterator affect the original Entity or not? (that's actually dependent on the implementation of `getProperty` -- whether it copies its internal representation or returns it directly)

e.g. could rewrite as

```java
if (potentialMatches != null && !potentialMatches.isEmpty()) {
  nextPotentialMatchID = potentialMatches.get(0);
  matchIntoEntity.setProperty(POTETNIAL_MATCHES_PROPERTY, potentialMatches.subList(1, potentialMatches.size()));
} else {
  // ...
}
```

This also avoids any assumption that the return value of `getProperty` is mutable -- e.g. if `getProperty` returns an `ImmutableList`, calling `iterator.remove()` would throw.

src/main/java/com/google/sps/servlets/PotentialMatchesServlet.java
```diff
@@ -91,7 +142,6 @@ private String loadUserPotentialMatch(String userID) {
   * @return the set of user nodes for all current app users
   */
   private ImmutableSet<UserNode> createUserNodes() {
```
daniel-google, line 144: nit: since this isn't actually creating anything in the datastore, I would rename to e.g. `getAllUserNodes`

(even though it's "creating" Java objects, verbs used in a servlet should generally describe the long-term effects, of which there are none here)

src/main/java/com/google/sps/servlets/PotentialMatchesServlet.java
```diff
@@ -60,28 +70,69 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
+    //Run the potential matching algorithm to find all potential matches
     ImmutableSet<String> potentialMatches = PotentialMatchAlgorithm.findPotentialMatchesForUser(userID, friendsMap);
-  
-    String nextPotentialMatchID = potentialMatches.iterator().hasNext() ?
-      potentialMatches.iterator().next() : NO_POTENTIAL_MATCH_RESULT;
+
+    //Add new Match Info entity to datastore
+    Entity newMatchInfo = new Entity(MATCH_INFO_ENTITY);
+    newMatchInfo.setProperty(USER_ID_PROPERTY, userID);
+    newMatchInfo.setProperty(POTENTIAL_MATCHES_PROPERTY, new ArrayList<String>(potentialMatches));
```
daniel-google, line 131:
```suggestion
    newMatchInfo.setProperty(POTENTIAL_MATCHES_PROPERTY, ImmutableList.copyOf(potentialMatches));
```

(prefer immutable types wherever you can)

src/main/java/com/google/sps/servlets/PotentialMatchesServlet.java
```diff
@@ -60,28 +70,69 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
+    newMatchInfo.setProperty(FRIENDED_IDS_PROPERTY, new ArrayList<String>());
+    newMatchInfo.setProperty(PASSED_IDS_PROPERTY, new ArrayList<String>());
```
daniel-google, line 133:
```suggestion
    newMatchInfo.setProperty(FRIENDED_IDS_PROPERTY, ImmutableList.of());
    newMatchInfo.setProperty(PASSED_IDS_PROPERTY, ImmutableList.of());
```

(prefer immutable types wherever you can)

src/test/java/com/google/sps/servlets/PotentialMatchesServletTest.java
```diff
@@ -199,14 +233,35 @@ public void fourUsersOneMutualConnection() throws Exception {
...
     return jsonResponse.getString(MATCHINFO_NEXT_MATCH_ID_FIELD);
   }
 
+  /**
+  * Checks if a user's match information was stored in datastore correctly
+  *
+  * @param currUserID The user ID of the user who's match information is getting checked
+  * @param expectedMatches The list of potential matches that are expected to be in datastore
+  */
+  private void datastoreAssertions(String currUserID, List<String> expectedMatches) {
```
daniel-google, line 255: nit: rename to what the method does, e.g. `assertUserMatchesInDatastore`

src/test/java/com/google/sps/servlets/PotentialMatchesServletTest.java
```diff
@@ -35,16 +37,27 @@
...
 
 @RunWith(JUnit4.class)
 public class PotentialMatchesServletTest {
   private static final String NO_POTENTIAL_MATCH_RESULT = "NO_POTENTIAL_MATCHES";
   private static final String MATCHINFO_NEXT_MATCH_ID_FIELD = "nextPotentialMatchID";
   private static final String USER_ID_REQUEST_URL_PARAM = "userid";
+  private static final String MATCH_INFO_ENTITY = "match-info";
+  private static final String POTENTIAL_MATCHES_PROPERTY = "potential-matches";
+  private static final String FRIENDED_IDS_PROPERTY = "friended-ids";
+  private static final String PASSED_IDS_PROPERTY = "passed-ids";
```
daniel-google, line 60: any particular reason these are duplicated here?

pom.xml
mivdittamo: When I merged the main branch with this branch, some of the dependencies ended up getting duplicated, so I removed the duplicated ones here.

src/main/java/com/google/sps/servlets/PotentialMatchesServlet.java
mivdittamo: Got it! I went with the first option and changed this method to return an Entity.

src/main/java/com/google/sps/servlets/PotentialMatchesServlet.java
mivdittamo: I changed the method name to advanceToNextPotentialMatch and also updated the javadocs description to mention that the method is also removing the next ID from the user's potential matches list.

src/main/java/com/google/sps/servlets/PotentialMatchesServlet.java
mivdittamo: Got it!

src/main/java/com/google/sps/servlets/PotentialMatchesServlet.java
mivdittamo: That makes sense. I changed this to include your rewrite suggestion!

src/main/java/com/google/sps/servlets/PotentialMatchesServlet.java
mivdittamo: Got it!

src/test/java/com/google/sps/servlets/PotentialMatchesServletTest.java
mivdittamo: Got it!

src/main/java/com/google/sps/servlets/PotentialMatchesServlet.java
mivdittamo: Got it!

src/main/java/com/google/sps/servlets/PotentialMatchesServlet.java
mivdittamo: Got it!

src/test/java/com/google/sps/servlets/PotentialMatchesServletTest.java
mivdittamo: The constants on line 54 through 58 are being used in the test cases to refer to certain parameter, entity, and property names and certain values. I'm noticing that I didn't use the last two properties, so I can remove these.

daniel-google: Yeah, what I meant was that you could reference the constants already defined in PotentialMatchesServlet to avoid duplicating (after making them package-private).

Though sometimes tests *intentionally* duplicate values that should never be changed, so that the test will detect if someone *does* change them. These strings fall into that category, so what you have is also fine.

src/main/java/com/google/sps/servlets/PotentialMatchesServlet.java
```diff
@@ -60,46 +70,92 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
+  * @return The datastore entity of the user's match information
+  */
+  private Entity loadUserMatchInformation(String userID) {
+    Entity matchInfoEntity = datastore.prepare(new Query(MATCH_INFO_ENTITY).setFilter(
+      new FilterPredicate(USER_ID_PROPERTY, FilterOperator.EQUAL, userID))).asSingleEntity();
+    
+    if (matchInfoEntity == null) {
+      return addMatchInfoToDatastore(userID);
+    }
+    return matchInfoEntity;
```
daniel-google, line 88: supernit: this would be ever-so-slightly easier to read as

```java
if (matchInfoEntity != null) {
  return matchInfoEntity;
}
return addMatchInfoToDatastore(userId);
```

(acts on `matchInfoEntity` immediately after the null check, so that the reader doesn't need to think about that variable anymore when reading forward)

src/test/java/com/google/sps/servlets/PotentialMatchesServletTest.java
```diff
@@ -186,9 +206,22 @@ public void fourUsersOneMutualConnection() throws Exception {
     addTestUserEntityToDatastore(datastore, TEST_USER_4_ID, TEST_USER_4_NAME,
       TEST_USER_4_EMAIL, TEST_USER_4_BIO, testUser4FriendsList);
 
-    String actualOutput = execute(TEST_USER_4_ID);
+    String actualOutput_1 = execute(TEST_USER_4_ID);
```
daniel-google, line 209: nit: `actualOutput1` (usually underscores are not combined with camelCase, but if you've seen this elsewhere or Nik recommended it, feel free to leave as-is)

(also applies to other `*Output` variables)

src/main/java/com/google/sps/servlets/PotentialMatchesServlet.java
mivdittamo: Got it!

src/test/java/com/google/sps/servlets/PotentialMatchesServletTest.java
mivdittamo: Got it!

PR #29, opened 2020-07-22
Store images using blobstore, store blobkeys in datastore, and load images on profile page

jessdejong:

Three things were accomplished: storing images using Blobstore, storing BlobKeys in Datastore, and loading images on the profile page (upon page reload).

### Is this fixing an issue?

none

How it works: Use blobstore to get a url to upload images. Submit the user profile form to this url, to save images in blobstore. This is then redirected to the user-data servlet, where blobkeys are retrieved and stored in datastore. When an image is to be displayed, perform a get request to the BlobServlet.java to 'serve' the image. This allows the javascript to get a URL (to the image) to set an image tag.

- [ x ] Test written/updating
- [ x ] Tests passing


src/main/java/com/google/sps/servlets/UserDataServlet.java
```diff
@@ -118,12 +152,39 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
   }
 
+  /** Method that stores the blob-keys (in Datastore) of files uploaded to Blobstore */
+  private void getAndStoreBlobKeys(HttpServletRequest request, Entity userEntity) {
+    // Get values determining whether or not to update/save blobkeys of any particular image
+    String profilePhotoUploaded = getStringParameter(request, USER_PROFILE_PHOTO, IMAGE_NOT_FOUND);
+    String photo2Uploaded = getStringParameter(request, USER_PHOTO_2, IMAGE_NOT_FOUND);
+    String photo3Uploaded = getStringParameter(request, USER_PHOTO_3, IMAGE_NOT_FOUND);
+    String photo4Uploaded = getStringParameter(request, USER_PHOTO_4, IMAGE_NOT_FOUND);
+    String photo5Uploaded = getStringParameter(request, USER_PHOTO_5, IMAGE_NOT_FOUND);
```
mivdittamo, line 169: It seems like here that these will only ever have the string values of "true" or "false". Given that, is there a way to do this using boolean values instead of strings?

jessdejong:
@estep-google @daniel-google ready to go!

src/main/java/com/google/sps/servlets/UserDataServlet.java
```diff
@@ -154,5 +230,36 @@ private String getStringParameter(HttpServletRequest request, String name, Strin
...
+    BlobKey blobKey = blobKeys.get(0);
+
+    // User submitted form without selecting a file, so we can't get a URL. (live server)
+    if (!testing) {
+      BlobInfo blobInfo = new BlobInfoFactory().loadBlobInfo(blobKey);
+      if (blobInfo.getSize() == 0) {
+        blobstore.delete(blobKey);
+        return null;
+      }
+    }
```
daniel-google, line 261: Why do we need to branch this behavior for testing?

src/main/java/com/google/sps/servlets/UserDataServlet.java
```diff
@@ -51,7 +62,19 @@
...
+  private static final String USER_PHOTO_3 = "photo-3";
+  private static final String USER_PHOTO_4 = "photo-4";
+  private static final String USER_PHOTO_5 = "photo-5";
+  private static final String PROFILE_PHOTO_BLOBKEY = "profile-photo-blobkey";
+  private static final String PHOTO_2_BLOBKEY = "photo-2-blobkey";
+  private static final String PHOTO_3_BLOBKEY = "photo-3-blobkey";
+  private static final String PHOTO_4_BLOBKEY = "photo-4-blobkey";
+  private static final String PHOTO_5_BLOBKEY = "photo-5-blobkey";
 
+  private boolean testing = false;
```
daniel-google, line 76: Generally, branching behavior for testing is a bad practice as it makes code harder to reason about (tests diverge from production). If there is some functionality that doesn't work in a test environment, prefer to use a mock/fake to override that behavior.

src/main/java/com/google/sps/servlets/UserDataServlet.java
```diff
@@ -118,12 +151,39 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
       }
     }
+    getAndStoreBlobKeys(request, userEntity);
     datastore.put(userEntity);
 
     // Redirect to the profile page, and let the front-end know the current logged in user
     response.sendRedirect("/profile.html?id=" + userId);
   }
 
+  /** Method that stores the blob-keys (in Datastore) of files uploaded to Blobstore */
```
daniel-google, line 161:
```suggestion
  /** Stores the blob-keys (in Datastore) of files uploaded to Blobstore. */
```

src/main/java/com/google/sps/servlets/UserDataServlet.java
```diff
@@ -118,12 +151,39 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
+  private void getAndStoreBlobKeys(HttpServletRequest request, Entity userEntity) {
+    // Get values determining whether or not to update/save blobkeys of any particular image
+    boolean profilePhotoUploaded = getBooleanParameter(request, USER_PROFILE_PHOTO, IMAGE_NOT_FOUND);
```
daniel-google, line 164: nit: I would just call this USER_PHOTO_1 for consistency; the fact that USER_PHOTO_1 is used as the profile photo isn't relevant when it comes to storage

src/main/java/com/google/sps/servlets/UserDataServlet.java
```diff
@@ -51,7 +62,19 @@
   private static final String USER_FRIENDS_LIST_PROPERTY = "friends-list";
   private static final String USER_ID_PROPERTY = "id";
   private static final String USER_NAME_PROPERTY = "name";
+  private static final String USER_PROFILE_PHOTO = "profile-photo";
+  private static final String USER_PHOTO_2 = "photo-2";
+  private static final String USER_PHOTO_3 = "photo-3";
+  private static final String USER_PHOTO_4 = "photo-4";
+  private static final String USER_PHOTO_5 = "photo-5";
```
daniel-google, line 69: Can we store these as a `List<>` instead of N hard-coded fields?

We can still enforce a limit on the list size, but it would reduce duplication and make it easier to change the number of images in the future.

src/main/java/com/google/sps/servlets/UserDataServlet.java
```diff
@@ -136,6 +196,22 @@ void doPostWrapper(DatastoreService datastoreService, HttpServletRequest request
...
+    datastore = datastoreService;
+    doPost(request, response);
+  }
+
+  /** Wrap the getUploadedFileBlobKey method to test how Blobstore results are handled, for jUnit testing */
+  String getUploadedFileBlobKeyWrapper(HttpServletRequest request, String formInputElementName, BlobstoreService blobstoreService, boolean isTesting) {
+    blobstore = blobstoreService;
+    testing = isTesting;
+    return getUploadedFileBlobKey(request, formInputElementName);
+  }
```
daniel-google, line 213: These methods, as well as the doGet/doPostWrapper methods preceding them, make it very difficult to reason about how behavior differs between tests/production.

In general, avoid putting any test-specific code alongside production code. If there is special setup that needs to be done for tests, it should happen in the test itself, and this class can expose whatever is necessary to accommodate that (e.g. make the `blobstore`/`datastore` fields package-private instead of private, and override them with mocks in your jUnit `#setUp` method).

src/main/java/com/google/sps/servlets/UserDataServlet.java
```diff
@@ -154,5 +230,36 @@ private String getStringParameter(HttpServletRequest request, String name, Strin
     String[] value = request.getParameterValues(name);
     return value == null ? defaultValue : value;
   }
+
+  /** Returns the request parameter (for booleans), or the default value if not specified */
+  private boolean getBooleanParameter(HttpServletRequest request, String name, boolean defaultValue) {
+    String value = request.getParameter(name);
+    if (value == null) return false;
+    else return value.equals("true");
```
daniel-google, line 238: nit: `return Boolean.valueOf(request.getParameter(name));`

src/main/java/com/google/sps/servlets/UserDataServlet.java
```diff
@@ -154,5 +230,36 @@ private String getStringParameter(HttpServletRequest request, String name, Strin
     String[] value = request.getParameterValues(name);
     return value == null ? defaultValue : value;
   }
+
+  /** Returns the request parameter (for booleans), or the default value if not specified */
+  private boolean getBooleanParameter(HttpServletRequest request, String name, boolean defaultValue) {
```
daniel-google, line 235: nit: `defaultValue` seems to be unused?

src/main/webapp/script.js
```diff
@@ -13,11 +13,17 @@
 // limitations under the License.
 
 const NO_MATCH =  "NO_POTENTIAL_MATCHES"; // must be kept in sync with the PotentialMatches.java servlet
+const PROFILE_PHOTO_BLOBKEY = "profile-photo-blobkey";
+const PHOTO_2_BLOBKEY = "photo-2-blobkey";
+const PHOTO_3_BLOBKEY = "photo-3-blobkey";
+const PHOTO_4_BLOBKEY = "photo-4-blobkey";
+const PHOTO_5_BLOBKEY = "photo-5-blobkey";
```
daniel-google, line 20: Would be good to replace hard-coded keys with a numeric limit in JS code, too.

src/test/java/com/google/sps/servlets/UserDataServletTest.java
```diff
@@ -47,13 +53,23 @@
 /** Tests the User Data Servlet */
 @RunWith(JUnit4.class)
 public final class UserDataServletTest {
+  private static final String PROFILE_PHOTO_BLOBKEY = "profile-photo-blobkey";
```
daniel-google, line 56: nit: can you add a comment above all these strings to clarify why they are duplicated in the test? vs. referencing the constants in production code

src/main/java/com/google/sps/servlets/BlobstoreURLServlet.java
```diff
@@ -0,0 +1,46 @@
...
+* Servlet that handles requests for a blobstore url for users to
+* upload files to.
+*/
+@WebServlet("/blobstore-url")
+public class BlobstoreURLServlet extends HttpServlet{
+  private final BlobstoreService blobstoreService = BlobstoreServiceFactory.getBlobstoreService();
+
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    response.setContentType("text/html");
```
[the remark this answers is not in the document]
jessdejong: `text/plain`?

src/main/java/com/google/sps/servlets/UserDataServlet.java
jessdejong: I couldn't figure out another way to "get around" this snippet of code. I couldn't figure out how to mock the `BlobInfo blobInfo = new BlobInfoFactory().loadBlobInfo(blobKey);` line. Do you have a suggestion on how I would test the method while including that snippet?

src/main/java/com/google/sps/servlets/UserDataServlet.java
jessdejong: Changed it to a list!

src/test/java/com/google/sps/servlets/UserDataServletTest.java
jessdejong: How would I use the constants from production code. That seems like it would be better. I tried importing the constants.

src/main/java/com/google/sps/servlets/UserDataServlet.java
jessdejong: Okay so it turns out that code snippet wasn't necessary, I removed it.

jessdejong:
@daniel-google I addressed most of your comments, and left like one or two questions. Also, I believe that because I committed one of your code suggestions you became one of the contributors to this pull request, so you'll have to follow the instructions that the github google CLA bot provided to sign that agreement I think.

src/main/java/com/google/sps/servlets/BlobstoreURLServlet.java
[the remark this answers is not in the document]
daniel-google: Hmm I don't know why that would be, but yeah, fine to leave as-is.

src/test/java/com/google/sps/servlets/UserDataServletTest.java
daniel-google: Yep, exactly.

src/main/java/com/google/sps/servlets/UserDataServlet.java
```diff
@@ -76,7 +93,8 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
           .put(USER_EMAIL_PROPERTY, userEntity.getProperty(USER_EMAIL_PROPERTY))
           .put(USER_FRIENDS_LIST_PROPERTY, (ArrayList<String>) userEntity.getProperty(USER_FRIENDS_LIST_PROPERTY))
           .put(USER_ID_PROPERTY, userEntity.getProperty(USER_ID_PROPERTY))
-          .put(USER_NAME_PROPERTY, userEntity.getProperty(USER_NAME_PROPERTY));
+          .put(USER_NAME_PROPERTY, userEntity.getProperty(USER_NAME_PROPERTY))
+          .put("blobkeys", (ArrayList<String>) userEntity.getProperty("blobkeys"));
```
daniel-google, line 97: "blobkeys" should be a constant like the other properties are

src/main/java/com/google/sps/servlets/UserDataServlet.java
```diff
@@ -154,5 +190,24 @@ private String getStringParameter(HttpServletRequest request, String name, Strin
     String[] value = request.getParameterValues(name);
     return value == null ? defaultValue : value;
   }
-}
 
+  /** Returns the request parameter (for booleans), or the default value if not specified */
```
daniel-google, line 194: nit: "default value" part of this comment is out-of-date

src/main/java/com/google/sps/servlets/UserDataServlet.java
```diff
@@ -118,22 +137,39 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
-    doGet(request, response);
-  }
+  /** Stores the blob-keys (in Datastore) of files uploaded to Blobstore. */
+  private void getAndStoreBlobKeys(HttpServletRequest request, Entity userEntity) {
+    // Get values determining whether or not to update/save blobkeys of any particular image
+    boolean profilePhotoUploaded = getBooleanParameter(request, USER_PHOTO_1_PROPERTY);
+    boolean photo2Uploaded = getBooleanParameter(request, USER_PHOTO_2_PROPERTY);
+    boolean photo3Uploaded = getBooleanParameter(request, USER_PHOTO_3_PROPERTY);
+    boolean photo4Uploaded = getBooleanParameter(request, USER_PHOTO_4_PROPERTY);
+    boolean photo5Uploaded = getBooleanParameter(request, USER_PHOTO_5_PROPERTY);
```
daniel-google, line 154: nit: i would just inline these in the if() statements

src/main/java/com/google/sps/servlets/UserDataServlet.java
```diff
@@ -118,22 +137,39 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
 
-  /** Wrap the doGet method for jUnit testing */
-  void doGetWrapper(DatastoreService datastoreService, HttpServletRequest request, HttpServletResponse response) throws IOException {
-    datastore = datastoreService;
-    doGet(request, response);
-  }
+  /** Stores the blob-keys (in Datastore) of files uploaded to Blobstore. */
+  private void getAndStoreBlobKeys(HttpServletRequest request, Entity userEntity) {
+    // Get values determining whether or not to update/save blobkeys of any particular image
+    boolean profilePhotoUploaded = getBooleanParameter(request, USER_PHOTO_1_PROPERTY);
```
daniel-google, line 150: nit: photo1Uploaded

src/main/java/com/google/sps/servlets/UserDataServlet.java
```diff
@@ -46,13 +56,20 @@
...
-
-  private DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+  private static final String USER_PHOTO_1_PROPERTY = "profile-photo";
+  private static final String USER_PHOTO_2_PROPERTY = "photo-2";
+  private static final String USER_PHOTO_3_PROPERTY = "photo-3";
+  private static final String USER_PHOTO_4_PROPERTY = "photo-4";
+  private static final String USER_PHOTO_5_PROPERTY = "photo-5";
+
+  BlobstoreService blobstore = BlobstoreServiceFactory.getBlobstoreService();
+  DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
```
daniel-google, line 72: nit: add a @VisibleForTesting annotation to each of these fields

(or if that annotation isn't available, just add a comment noting that they are visible only for testing)

src/test/java/com/google/sps/servlets/UserDataServletTest.java
```diff
@@ -102,7 +120,7 @@ public void testGetMethodWithInvalidId() throws Exception {
     when(mockResponse.getWriter()).thenReturn(writer);
 
     // Test the doGet method by calling the wrapper method
```
daniel-google, line 122: nit: update comment along with all other instances in this file (should no longer refer to "wrapper")

src/main/java/com/google/sps/servlets/UserDataServlet.java
```diff
@@ -118,22 +137,39 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
+    boolean photo2Uploaded = getBooleanParameter(request, USER_PHOTO_2_PROPERTY);
+    boolean photo3Uploaded = getBooleanParameter(request, USER_PHOTO_3_PROPERTY);
+    boolean photo4Uploaded = getBooleanParameter(request, USER_PHOTO_4_PROPERTY);
+    boolean photo5Uploaded = getBooleanParameter(request, USER_PHOTO_5_PROPERTY);
 
-  /** Wrap the doPost method for jUnit testing */
-  void doPostWrapper(DatastoreService datastoreService, HttpServletRequest request, HttpServletResponse response) throws IOException {
-    datastore = datastoreService;
-    doPost(request, response);
+    List<String> blobKeys = (ArrayList<String>) userEntity.getProperty("blobkeys");
```
daniel-google, line 156: nit: cast to List intead of ArrayList

jessdejong:
@googlebot I fixed it.

PR #30, opened 2020-07-22
Match functions

alikew24:

Write functions to display a users matches on their match page
### Is this fixing an issue?

none

none

- [  ] Test written/updating
- [ x ] Tests passing

Tested manually using a mock servlet


src/main/webapp/script.js
```diff
@@ -168,6 +170,60 @@ function loadProfile() {
   });
 }
 
+//matches.html functions
+function displayMatches() {
+  let id = getCurrentUserId();
+  if (id === null) {
+      return;
+  }
+  fetch('/matches?id=' + id).then(response => response.json()).then((matches) => {
```
mivdittamo, line 179: nit: name the request parameter `userid` for consistent naming since some other functions that pass in the id are using this as the name

src/main/webapp/script.js
```diff
@@ -168,6 +170,60 @@ function loadProfile() {
...
+    console.log(matches);
+    for (let i = 0; i < matches.length; i++) {
+      createCardElement(matches[i]);
+    }
+  });
+}
+
+function createCardElement(userID) {
```
mivdittamo, line 187: It seems that the `createCardElement` function is both creating the card element and adding it to the match container, but the naming just implies that an element is created.

 It may help to split up this functionality here so that this function returns that card element being created, and then the `displayMatches` function can add that element to the matches container to be displayed.

alikew24:
thanks! @daniel-google @estep-google

src/main/webapp/script.js
```diff
@@ -65,11 +65,12 @@ function displayPotentialMatchInfo(pmID) {
     let carouselContainer = document.getElementById("carousel-inner");
     let numPhotos = 5; //hard coded value for now, once blobstore works, this will be the amount of images the user has uploaded
```
estep-google, line 66: Use `const` instead of `let` where you can. It looks to me like `numPhotos` won't be updated later, so it can be `const`. There is at least one other place where this is also true.

src/main/webapp/script.js
```diff
@@ -65,11 +65,12 @@ function displayPotentialMatchInfo(pmID) {
...
     for (let i = 0; i < numPhotos; i++) {
+      let slideshowElement;
```
estep-google, line 68: For reference only - Use of `let` here is good because you are not assigning at instantiation time, but later.

src/main/webapp/script.js
```diff
@@ -154,12 +156,12 @@ function passButtonPressed() {
...
 function loadProfile() {
-  let id = getCurrentUserId();
-  if (id === null) {
+  let userID = getCurrentUserId();
+  if (userID === null) {
     return;
   }
-  console.log(id);
-  fetch('/user-data?id=' + id).then(response => response.json()).then((userinfo) => {
+  console.log(userID);
```
estep-google, line 163: Do you still need to log? If yes, two things:

1.  Generally you're better off including context here, not just a number (e.g., `"userID = " + userID`)
1.  Open an issue to look for and remove log statements - you don't want these in the final product

src/main/webapp/script.js
```diff
@@ -168,6 +170,60 @@ function loadProfile() {
   });
 }
 
+//matches.html functions
```
estep-google, line 173: I don't understand the purpose of this comment

src/main/webapp/script.js
```diff
@@ -168,6 +170,60 @@ function loadProfile() {
...
+function displayMatches() {
+  let userID = getCurrentUserId();
+  if (userID === null) {
+      return;
+  }
+  fetch('/matches?id=' + userID).then(response => response.json()).then((matches) => {
+    console.log(matches);
+    let matchContainer = document.getElementById('matches-container');
```
estep-google, line 181: You'll have to try this to see if I'm right, but I think this can also be `const` because you're not changing the memory address of the variable itself.

src/main/webapp/script.js
```diff
@@ -168,6 +170,60 @@ function loadProfile() {
...
+    }
+  });
+}
+
+function createCardElement(userID) {
+  fetch('/user-data?id=' + userID).then(response => response.json()).then((userinfo) => {
+    if (userinfo === null) {
+        return;
+    }
+    var cardDiv = document.createElement("div");
```
estep-google, line 193: `var` is "old school" and discouraged ("banned" is too strong) in new code. You should use `let` or `const`.

See here for some details on `var`: https://javascript.info/var

I would like to not introduce them here in this PR, but let's no go crazy and fix any that are already committed. That would be a nice follow up, though.

PR #35, opened 2020-07-27
Add servlet to retrieve matches

mivdittamo:

- Add a servlet that returns the list of a user's current matches

- Add test cases to test the functionality of the matches servlet

### Is this fixing an issue?

none

none

- [ x ] Test written/updating
- [ x ] Tests passing

### DIFF
--- a/src/main/java/com/google/sps/servlets/MatchesListServlet.java
+++ b/src/main/java/com/google/sps/servlets/MatchesListServlet.java
@@ -44,7 +44,6 @@ public class MatchesListServlet extends HttpServlet {
   private static final String FRIENDED_IDS_PROPERTY = "friended-ids";
   private static final String PASSED_IDS_PROPERTY = "passed-ids";
   private static final String MATCHES_LIST_PROPERTY = "matches-list";
-
   private static final String USER_ID_REQUEST_URL_PARAM = "userid";
   
   DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
--- a/src/main/java/com/google/sps/servlets/MatchesListServlet.java
+++ b/src/main/java/com/google/sps/servlets/MatchesListServlet.java
@@ -1,0 +1,71 @@
+// Copyright 2019 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.sps.servlets;
+
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import com.google.gson.Gson;
+import com.google.common.collect.ImmutableList;
+import com.google.appengine.api.datastore.DatastoreService;
+import com.google.appengine.api.datastore.DatastoreServiceFactory;
+import com.google.appengine.api.datastore.Query;
+import com.google.appengine.api.datastore.PreparedQuery;
+import com.google.appengine.api.datastore.FetchOptions;
+import com.google.appengine.api.datastore.Query.FilterOperator;
+import com.google.appengine.api.datastore.Query.FilterPredicate;
+import com.google.appengine.api.datastore.Entity;
+
+/**
+ * Servlet that handles requests to retrieve a list of a user's matches
+ */
+@WebServlet("/match-information")
+public class MatchesListServlet extends HttpServlet {
+  private static final String MATCH_INFO_ENTITY = "match-info";
+  private static final String USER_ID_PROPERTY = "id";
+  private static final String POTENTIAL_MATCHES_PROPERTY = "potential-matches";
+  private static final String FRIENDED_IDS_PROPERTY = "friended-ids";
+  private static final String PASSED_IDS_PROPERTY = "passed-ids";
+  private static final String MATCHES_LIST_PROPERTY = "matches-list";
+
+  private static final String USER_ID_REQUEST_URL_PARAM = "userid";
+  
+  DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    String userID = (String) request.getParameter(USER_ID_REQUEST_URL_PARAM);
+    
+    Entity matchInfoEntity = datastore.prepare(new Query(MATCH_INFO_ENTITY).setFilter(
+      new FilterPredicate(USER_ID_PROPERTY, FilterOperator.EQUAL, userID))).asSingleEntity();
+
+    List<String> matchedUsers = (List<String>) matchInfoEntity.getProperty(MATCHES_LIST_PROPERTY);
+    if (matchedUsers == null) {
+      matchedUsers = Arrays.asList();
+    }
+
+    Gson gson = new Gson();
+    String json = gson.toJson(matchedUsers);
+
+    response.setContentType("application/json");
+    response.getWriter().print(json);
+  }
+}
+
--- a/src/main/java/com/google/sps/servlets/MatchesListServlet.java
+++ b/src/main/java/com/google/sps/servlets/MatchesListServlet.java
@@ -36,7 +36,7 @@ import com.google.appengine.api.datastore.Entity;
 /**
  * Servlet that handles requests to retrieve a list of a user's matches
  */
-@WebServlet("/match-information")
+@WebServlet("/matches-list")
 public class MatchesListServlet extends HttpServlet {
   private static final String MATCH_INFO_ENTITY = "match-info";
   private static final String USER_ID_PROPERTY = "id";
--- a/src/test/java/com/google/sps/servlets/MatchesListServletTest.java
+++ b/src/test/java/com/google/sps/servlets/MatchesListServletTest.java
@@ -99,11 +99,10 @@ public class MatchesListServletTest {
     addMatchInfoEntityToDatastore(TEST_USER_ID, ImmutableList.of(),
       ImmutableList.of(), ImmutableList.of(), ImmutableList.of());
     
-    String actualOutput = execute(TEST_USER_ID);
-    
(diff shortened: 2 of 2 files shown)


src/main/java/com/google/sps/servlets/MatchesListServlet.java
```diff
@@ -0,0 +1,71 @@
...
+ */
+@WebServlet("/matches-list")
+public class MatchesListServlet extends HttpServlet {
+  private static final String MATCH_INFO_ENTITY = "match-info";
+  private static final String USER_ID_PROPERTY = "id";
+  private static final String POTENTIAL_MATCHES_PROPERTY = "potential-matches";
+  private static final String FRIENDED_IDS_PROPERTY = "friended-ids";
+  private static final String PASSED_IDS_PROPERTY = "passed-ids";
+  private static final String MATCHES_LIST_PROPERTY = "matches-list";
+
```
alikew24, line 47: nit: i am not sure if this line new line is needed here

src/test/java/com/google/sps/servlets/MatchesListServletTest.java
```diff
@@ -0,0 +1,174 @@
...
+
+@RunWith(JUnit4.class)
+public class MatchesListServletTest {
+  private static final String MATCH_INFO_ENTITY = "match-info";
+  private static final String USER_ID_PROPERTY = "id";
+  private static final String POTENTIAL_MATCHES_PROPERTY = "potential-matches";
+  private static final String FRIENDED_IDS_PROPERTY = "friended-ids";
+  private static final String PASSED_IDS_PROPERTY = "passed-ids";
+  private static final String MATCHES_LIST_PROPERTY = "matches-list";
+
```
alikew24, line 53: nit: same here with the new lines- not sure exactly what the protocol is though so if they are needed then no worries

src/test/java/com/google/sps/servlets/MatchesListServletTest.java
```diff
@@ -0,0 +1,174 @@
...
+  private static final String USER_ID_REQUEST_URL_PARAM = "userid";
+
+  private static final String TEST_USER_ID = "5555";
```
alikew24, line 56: maybe add comments here explaining what these are

src/main/java/com/google/sps/servlets/MatchesListServlet.java
mivdittamo: Got it!

src/test/java/com/google/sps/servlets/MatchesListServletTest.java
```diff
@@ -0,0 +1,174 @@
...
+  @Test
+  public void matchInfoWithEmptyList() throws IOException{
+    addMatchInfoEntityToDatastore(TEST_USER_ID, ImmutableList.of(),
+      ImmutableList.of(), ImmutableList.of(), ImmutableList.of());
+    
+    String actualOutput = execute(TEST_USER_ID);
+    
+    List<String> actualOutputAsList = Arrays.asList(new Gson().fromJson(actualOutput, String[].class));
+
+    assertThat(actualOutputAsList).isEqualTo(EMPTY_LIST);
```
estep-google, line 106: You should just be able to do `assertThat(actualOutputAsList).isEmpty();`

src/test/java/com/google/sps/servlets/MatchesListServletTest.java
```diff
@@ -0,0 +1,174 @@
...
+   * Tests the case in which a user's match list is empty.
+   *
+   * <p>Should return an empty list.
+   */
+  @Test
+  public void matchInfoWithEmptyList() throws IOException{
+    addMatchInfoEntityToDatastore(TEST_USER_ID, ImmutableList.of(),
+      ImmutableList.of(), ImmutableList.of(), ImmutableList.of());
+    
+    String actualOutput = execute(TEST_USER_ID);
```
estep-google, line 102: Let's keep the naming/layout simple here:

```
String jsonOutput = execute(TEST_USER_ID);
List<String> matches = Arrays.asList(new Gson().fromJson(jsonOutput, String[].class));
```

Generally (though it's not a hard and fast rule), I like to group my lines of code for a test in three "blocks" where each "block" is separated:

```
@Test
public void niksAwesomeTest() throws Exception {
  testSetUpLineNo1;
  testSetUpLineNo2;
  testSetUpLineNo3;

  runTestLineNo1;
  runTestLineNo2;

  assertResultLineNo1;
  assertResultLineNo2;
  assertResultLineNo3;
}
```

src/test/java/com/google/sps/servlets/MatchesListServletTest.java
```diff
@@ -0,0 +1,174 @@
...
+    helper.tearDown();
+  }
+
+  /**
+   * Tests the case in which a user's match list is empty.
+   *
+   * <p>Should return an empty list.
+   */
+  @Test
+  public void matchInfoWithEmptyList() throws IOException{
```
estep-google, line 98: 1) For a test, you can just `throws Exception` (pretty much just do that with all tests always unless you have a good reason not to)
2) Make sure there's a space between `Exception` and `{`

src/test/java/com/google/sps/servlets/MatchesListServletTest.java
```diff
@@ -0,0 +1,174 @@
...
+   */
+  private String execute(String userIDToFetch) throws IOException{
+    when(mockRequest.getParameter(USER_ID_REQUEST_URL_PARAM)).thenReturn(userIDToFetch);
+    
+    responseWriter = new StringWriter();
+    writer = new PrintWriter(responseWriter, true);
+    when(mockResponse.getWriter()).thenReturn(writer);
+
+    servletUnderTest.doGet(mockRequest, mockResponse);
+    System.out.println(responseWriter);
```
estep-google, line 157: Remove

src/test/java/com/google/sps/servlets/MatchesListServletTest.java
```diff
@@ -0,0 +1,174 @@
...
+
+  /**
+   * Method that calls on the MatchInformationServlet and returns the ID of the next
+   * potential match for the specified user
+   *
+   * @param userIDToFetch The ID of the user whose next potential match is being found
+   */
+  private String execute(String userIDToFetch) throws IOException{
+    when(mockRequest.getParameter(USER_ID_REQUEST_URL_PARAM)).thenReturn(userIDToFetch);
+
```
estep-google, line 151: Remove.

Not as strictly a rule here, but I would read this method as "setup" -> "test" -> "assert" where you actually have "setup" -> "execute" -> "return". Grouping by blocks agin.

src/test/java/com/google/sps/servlets/MatchesListServletTest.java
```diff
@@ -0,0 +1,174 @@
...
+    servletUnderTest.doGet(mockRequest, mockResponse);
+    System.out.println(responseWriter);
+    
+    return responseWriter.toString();
+  }
+
+  private void addMatchInfoEntityToDatastore(String userID, ImmutableList<String> potentialMatches, 
+    ImmutableList<String> friendedUsers, ImmutableList<String> passedUsers, ImmutableList<String> matchedUsers) {
+
```
estep-google, line 164: Remove

src/test/java/com/google/sps/servlets/MatchesListServletTest.java
```diff
@@ -0,0 +1,174 @@
...
+  }
+
+  /**
+   * Tests the case in which a user's match list is empty.
+   *
+   * <p>Should return an empty list.
+   */
+  @Test
+  public void matchInfoWithEmptyList() throws IOException{
+    addMatchInfoEntityToDatastore(TEST_USER_ID, ImmutableList.of(),
```
estep-google, line 99: While I would *really really* like a builder here, it's overkill.

That said, we do need more context here on what these empty lists are. We achieve that this way:

```
addMatchInfoEntityToDatastore(TEST_USER_ID, /* param1Name= */ ImmutableList.of(), /* param2Name= */ ImmutableList.of(),....
```

User id is pretty clear here, but you need to label the other method parameters - they are just lists w/o inline context. Also, make sure you need all of the lists. I think I see at least the first one is always empty. If you're not setting it in the test, just set it to a default it in the `addMatchInfoEntityToDatastore` method itself.

src/test/java/com/google/sps/servlets/MatchesListServletTest.java
estep-google, line 98: In most cases, the name of the test and then the assertions in the test should make it clear what's being tested. I try to not need javadocs for test cases (not always possible). Here, I think it is, though.

* `hasNoMatches`
* `hasSingleMatch`
* `hasMultipleMatches`

Test names are very clear here and there's no need for javadoc.

src/test/java/com/google/sps/servlets/MatchesListServletTest.java
```diff
@@ -0,0 +1,174 @@
...
+    List<String> actualOutputAsList = Arrays.asList(new Gson().fromJson(actualOutput, String[].class));
+
+    assertThat(actualOutputAsList).containsExactly(TEST_CONNECTION_1_ID, TEST_CONNECTION_2_ID, TEST_CONNECTION_3_ID);
+  }
+
+  /**
+   * Method that calls on the MatchInformationServlet and returns the ID of the next
+   * potential match for the specified user
+   *
+   * @param userIDToFetch The ID of the user whose next potential match is being found
```
estep-google, line 147: If you're going to include an `@param`, you should include an `@returns`.

src/test/java/com/google/sps/servlets/MatchesListServletTest.java
mivdittamo: Got it!

src/test/java/com/google/sps/servlets/MatchesListServletTest.java
mivdittamo: Got it!

src/test/java/com/google/sps/servlets/MatchesListServletTest.java
mivdittamo: Got it!

src/test/java/com/google/sps/servlets/MatchesListServletTest.java
mivdittamo: got it!

src/test/java/com/google/sps/servlets/MatchesListServletTest.java
mivdittamo: Got it! I saw that the passed and potential match lists were always empty so I removed those parameters from addMatchInfoEntityToDatastore. And I kept the friendedUsers lists because I realized that users on the match list should also be in the friendedUsers list.

src/test/java/com/google/sps/servlets/MatchesListServletTest.java
mivdittamo: Got it!

src/test/java/com/google/sps/servlets/MatchesListServletTest.java
mivdittamo: got it!

PR #39, opened 2020-07-29
Load images from blobstore into Carousel and Matches Page

alikew24:

The Carousel and Matches Cards now load the images from Blobstore to display to the user
### Is this fixing an issue?

None

None

- [  ] Test written/updating
- [ x ] Tests passing

passing manual tests


src/main/webapp/script.js
```diff
@@ -58,16 +58,23 @@ function displayPotentialMatchInfo(pmID) {
...
     const bio = userinfo.bio;
     const carouselContainer = document.getElementById("carousel-inner");
-    const numPhotos = 5; //hard coded value for now, once blobstore works, this will be the amount of images the user has uploaded
-    for (let i = 0; i < numPhotos; i++) {
+    let numPhotos = 0;
+    for (let i = 0; i < userinfo.blobkeys.length; i++) {
       let slideshowElement;
-      if (i === 0) {
-        slideshowElement = createSlideshowElement("images/noBlobStoreImage.jpg", "carousel-item active", name, bio);
+      if (i === 0 && userinfo.blobkeys[i] != "") {
```
mivdittamo, line 64: Use !== instead of != for a more strict comparison

src/main/webapp/script.js
```diff
@@ -58,16 +58,23 @@ function displayPotentialMatchInfo(pmID) {
...
+        numPhotos++;
+        slideshowElement = createSlideshowElement(userinfo.blobkeys[i], "carousel-item active", name, bio, false);
+        carouselContainer.appendChild(slideshowElement);
       }
-      else {
-        slideshowElement = createSlideshowElement("images/noBlobStoreImage.jpg", "carousel-item", name, bio);
+      else if (userinfo.blobkeys[i] != "") {
```
mivdittamo, line 69: Same here, use !==

src/main/webapp/script.js
```diff
@@ -196,11 +210,23 @@ function displayMatches() {
...
+    console.log(userinfo);
     if (userinfo === null) {
         return;
     }
     cardDiv.className = "col card m-5 card-container";
-    const profileImage = createImgElement("images/noBlobStoreImage2.jpg");
+    let profileImage;
+    let foundImage = false;
+    for (let i = 0; i < userinfo.blobkeys.length; i++) {
+      if (userinfo.blobkeys[i] != "") {
```
mivdittamo, line 221: Same here, use !==

src/main/webapp/script.js
```diff
@@ -58,16 +58,23 @@ function displayPotentialMatchInfo(pmID) {
...
+        numPhotos++;
+        slideshowElement = createSlideshowElement(userinfo.blobkeys[i], "carousel-item active", name, bio, false);
+        carouselContainer.appendChild(slideshowElement);
       }
-      else {
-        slideshowElement = createSlideshowElement("images/noBlobStoreImage.jpg", "carousel-item", name, bio);
+      else if (userinfo.blobkeys[i] != "") {
+        numPhotos++;
+        slideshowElement = createSlideshowElement(userinfo.blobkeys[i], "carousel-item", name, bio, false);
+        carouselContainer.appendChild(slideshowElement);
```
daniel-google, line 72: Looks like the only difference between this and the preceding if() is the class name? If so, can simplify to

```js
if (userinfo.blobkeys[i] != "") {
  numPhotos++;
  slideshowElement = createSlideshowElement(userinfo.blobkeys[i], "carousel-item" + (i === 0 ? " active" : ""), name, bio, false);
  carouselContainer.appendChild(slideshowElement);
}
```

src/main/webapp/script.js
```diff
@@ -78,10 +85,15 @@ function deletePotentialMatchInfo() {
   document.getElementById("carousel-indicators").innerHTML = '';
 }
 
-function createSlideshowElement(blobkey, className, name, bio) {
+function createSlideshowElement(blobkey, className, name, bio, noMatch) {
```
daniel-google, line 88: nit: rather than passing in a boolean here, you could just pass in the image element, i.e. the caller would invoke either createImgElement or createImageFromBlobStore. Makes it a bit more clear what the parameter represents, and you'd also remove the blobKey parameter from this function (passing "images/no_image.png" as a blob key in the no-match case is a bit confusing, since that's not actually a blob key).

src/main/webapp/script.js
daniel-google, line 221: Question for my own understanding: is it ever *expected* that the server pass an empty string in blobKeys?

src/main/webapp/script.js
```diff
@@ -196,11 +210,23 @@ function displayMatches() {
...
-    const profileImage = createImgElement("images/noBlobStoreImage2.jpg");
+    let profileImage;
+    let foundImage = false;
+    for (let i = 0; i < userinfo.blobkeys.length; i++) {
+      if (userinfo.blobkeys[i] != "") {
+        foundImage = true;
+        profileImage = createImageFromBlobstore(userinfo.blobkeys[i]);
+        break;
+      }
+    }
```
daniel-google, line 226: I'd move this logic to its own function `getProfileImage` that returns either an image element, or a nullable blobKey.

src/main/webapp/script.js
```diff
@@ -281,3 +307,14 @@ function setImageFromBlobstore(imageBlobKey, imageId) {
     document.getElementById(imageId).src = URL.createObjectURL(blobContent);
   });
 }
+
+//Function that creates an image element that corresponds to the provided blobkey and returns it
+function createImageFromBlobstore(imageBlobKey) {
```
daniel-google, line 312: Hmm this function will create and return an `img` element *without* an `src` attribute, then populate it later on once the fetch complete. That feels a bit awkward, but probably fine from a functional perspective.

Is there a reason the servlet doesn't just return the image URL? Since it doesn't look like the client can do much with the `blobKey` alone.

src/main/webapp/script.js
alikew24: yeah- that's the default. When the user has an image in position 2 say, then bobkeys[2] will have an image path. Otherwise, it will be ""

src/main/webapp/script.js
alikew24: I'm pretty sure Isabelle had an issue in the walkthroughs with returning the url, not completely sure though. I think in all the tutorials the servlet returns the blobkey? Gotta check with Jess on this.

alikew24:
I addressed some of the comments and responded to the others! @daniel-google

src/main/webapp/script.js
```diff
@@ -58,16 +58,20 @@ function displayPotentialMatchInfo(pmID) {
     const name = userinfo.name;
     const bio = userinfo.bio;
     const carouselContainer = document.getElementById("carousel-inner");
-    const numPhotos = 5; //hard coded value for now, once blobstore works, this will be the amount of images the user has uploaded
-    for (let i = 0; i < numPhotos; i++) {
+    let numPhotos = 0;
+    for (let i = 0; i < userinfo.blobkeys.length; i++) {
       let slideshowElement;
```
daniel-google, line 63: nit: this can be moved to the same line which assigns it

src/main/webapp/script.js
```diff
@@ -232,6 +240,15 @@ function createCardElement(userID) {
   return cardDiv;
 }
 
+function getFirstAvailableImage(blobkeyList) {
+  for (let i = 0; i < blobkeyList.length; i++) {
+    if (blobkeyList[i] !== "") {
+      return(createImageFromBlobstore(userinfo.blobkeys[i]));
+      }
+    }
+    return(createImgElement("images/no_image.png"));
```
daniel-google, line 249: nit: fix indentation

daniel-google, line 249: nit: return isn't a function, remove parenthesis (just `return createImageFromBlobstore(...);`)

src/main/webapp/script.js
daniel-google: Okay, not anything that needs to be addressed in this CL, was just curious!

src/main/webapp/script.js
daniel-google: Got it, thanks!

